PR #23, opened 2023-01-08
Extend cygwin support to Windows XP SP3

JoshuaWierenga:
[Original windows branch](https://github.com/JoshuaWierenga/blink/tree/add_windows_support) got too complicated and was largely redundant after 74ca808bfe5c9ed8203e00d1bfe0e54ab2741c2b so broke out the remaining xp changes into a separate branch.

Includes Cygwin and MinGW's [weird decision](https://github.com/mirror/mingw-w64/blob/7a411b6ae3bb1e83ed932b493dac044e7c62be1c/mingw-w64-headers/crt/_mingw.h.in#L18-L27) to make LONG actually become 64 bit on x86_64 unlike Windows so I have to use __LONG32 to make things work properly.

### DIFF
--- a/blink/random.c
+++ b/blink/random.c
@@ -37,6 +37,12 @@
 #include "blink/errno.h"
 #include "blink/random.h"
 #include "blink/types.h"
+#if defined(__CYGWIN__)
+#include <stdbool.h>
+#include <w32api/_mingw.h>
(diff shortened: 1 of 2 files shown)


jart (approved):
Thanks!

Issue #8, opened 2022-12-04
Porting to webassembly


saolsen:
Hey,

I'm thinking about trying to get blink running in the browser (via webassembly). My goal isn't just to run c code in the browser (could just use emscripten for that) but to actually run ape x64 executables in an interpreter where I can build a debugger and some visualization tools to see what's happening. (Similar to what blink already does).
I'm wondering if anybody has tried to compile blink for webassembly or what you think some of the challenges would be.

jart:
Sounds awesome. I haven't tried it myself. I imagine the only major obstacle might be 32-bit. JavaScript only supports 32-bit integers. Blink runs on i386 and other 32-bit CPUs, and is regularly tested on them. However Blink was created in a 64-bit world, and as such, Blink has a 64-bit bias, therefore, you might not get optimal performance out of Blink in your 32-bit environments.

unicomp21:
https://github.com/WebAssembly/tail-call/issues/15#issuecomment-1357820841

We're still waiting on tail calls for webassembly, lol.  Wish there was another option.

unicomp21:
Someday we'll be able to use co_await in webassembly, hope I'm still around to see it happen, lol.

https://github.com/WebAssembly/tail-call/issues/14#issue-1058389687

unicomp21:
https://github.com/emscripten-core/emscripten/issues/10991#issuecomment-974226917

jart:
Blink doesn't need tail calls.

Also, good news! Blink is now stable on 32-bit platforms. I don't know if webassembly is multi-core, but the threading issues Blink was having earlier on 32-bit have been resolved.

Your biggest obstacle is most likely going to be completely replacing everything in blink/syscall.c so it interfaces with WASI or something similar instead.

unicomp21:
https://github.com/copy/v86

unicomp21:
@gornishanov @tlively @madmongo1 @tomoliv any ideas on easiest way to implement linux syscall.c for webassembly?

https://github.com/jart/blink/issues/8#issuecomment-1370735621

https://github.com/WebKit/WebKit/pull/2065

unicomp21:
Dumb/crazy question, what sort of perf hit if blink runs nested within itself?  I'm wondering if the outer vm could implement syscall.c and handle threading etc. using co_await?

jart:
[...]

PR #25, opened 2023-01-13
WebAssembly GitHub Pages build

trungnt2910:
Upstreaming my branch (itself a fork of @Vogtinator's work) according to @jart's [request](https://github.com/jart/blink/issues/8#issuecomment-1381880230).

This branch is a WIP, with known issues:
- ~~Memory leak, see more details [here](https://github.com/jart/blink/issues/8#issuecomment-1381908007).~~
- ~~Somehow the sample binary (a cosmopolitan one) works, but a statically linked Hello World Linux binary (obtained by following the guide [here](https://jameshfisher.com/2018/03/10/linux-assembly-hello-world/)) doesn't work. See the error output [here](https://github.com/jart/blink/issues/8#issuecomment-1381977865).~~


.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,50 @@
...
+          curl -L https://github.com/emscripten-core/emsdk/archive/HEAD.tar.gz | tar xz
+          mv emsdk-* emsdk
+          ./emsdk/emsdk install latest
+          ./emsdk/emsdk activate latest
+
+      - name: Build blink
+        run: |
+          . ./emsdk/emsdk_env.sh
+          emmake make -j$(nproc)
+          curl -sSL https://raw.githubusercontent.com/gzuidhof/coi-serviceworker/master/coi-serviceworker.min.js -o o//blink/blink-web/coi-serviceworker.min.js --create-dirs
```
jart, line 35: Is it possible for this to run without installing a service worker? If not, then this code should be vendored in the third_party folder.

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,50 @@
...
+        id: cache
+        uses: actions/cache@v3
+        with:
+          path: 'emsdk'
+          key: 'emsdk'
+
+      - name: Install emscripten
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          curl -L https://github.com/emscripten-core/emsdk/archive/HEAD.tar.gz | tar xz
```
jart, line 26: Please use a URL that has a release version number.

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,50 @@
...
+            -s INITIAL_MEMORY=1073741824 \
+            -s EXIT_RUNTIME=1 \
+            -s EXPORTED_FUNCTIONS=_main \
+            -s EXPORTED_RUNTIME_METHODS=setValue,getValue,stringToUTF8Array,lengthBytesUTF8,cwrap \
+            -fno-builtin-exit \
+            -sASSERTIONS
+
+      - name: Deploy sample to GitHub pages
+        uses: JamesIves/github-pages-deploy-action@v4
+        # if: ${{ github.ref == 'refs/heads/master' }}
```
jart, line 48: Please pin a version here too if possible.

jart (changes requested):

blink/blink-shell.html
```diff
@@ -0,0 +1,276 @@
...
+        var size = Module.lengthBytesUTF8(s) + 1;
+        var ptr = Module._malloc(size);
+        Module.stringToUTF8Array(s, Module.HEAP8, ptr, size);
+        return ptr;
+      }
+
+      function runExecutable() {
+        let args = ['blink', 'executable'].concat(document.getElementById('argsInput').value.split(' '));
+
+        let main = Module.cwrap('main', 'number', ['number', 'number']);
```
Vogtinator, line 248: You can just export `callMain` instead and do `Module.callMain(args);'

blink/memorymalloc.c
```diff
@@ -67,7 +67,9 @@ static size_t GetBigSize(size_t n) {
 
 void FreeBig(void *p, size_t n) {
   if (!p) return;
+#if !defined(__EMSCRIPTEN__)
```
Vogtinator, line 70: This probably just needs a change to mirror the special behaviour in `AllocateBig` to just use the size as-is, i.e.

```
#if defined(__CYGWIN__) || defined(__EMSCRIPTEN__)
  unassert(!munmap(p, n));
#else
  unassert(!munmap(p, GetBigSize(n)));
#endif
```

blink/blink.c
```diff
@@ -187,6 +187,12 @@ static void HandleSigs(void) {
 #endif
 }
 
+#if defined(__EMSCRIPTEN__)
+void exit(int status) {
+  EM_ASM({ throw new ExitStatus(status); });
```
Vogtinator, line 192: Is `-s EXIT_RUNTIME=0` not enough?

.github/workflows/emscripten.yml
trungnt2910: Technically the service worker is _not_ required. However, in order to host this page on _GitHub_ Pages, this service worker has to be included to enable [cross-origin isolation](https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated).

This is required for pages that want to access `SharedArrayBuffer` (required for WASM threads), but GitHub Pages does not enable it by default.

.github/workflows/emscripten.yml
trungnt2910: You mean pinning a specific commit?

blink/blink-shell.html
trungnt2910: Doesn't `callMain` twice have some undesired effects?

blink/blink.c
trungnt2910: `EXIT_RUNTIME=0` (the default) makes calling `exit` does nothing but printing something to `stderr`.
`EXIT_RUNTIME=1` (the thing you did) makes calling `exit` prevents `main` from ever being called again.

.github/workflows/emscripten.yml
trungnt2910: Done. Pinned to v4.4.1

.github/workflows/emscripten.yml
trungnt2910: Done. Pinned to v3.1.30

blink/blink-shell.html
trungnt2910: Done.

.github/workflows/emscripten.yml
jart: Good to know. This is fine for now. I'm going to look into providing a `make MODE=nothreads` build mode so we can avoid installing service workers when we don't need them.

trungnt2910: Should I edit the Makefile to copy required scripts into the output folder or is doing this from the GitHub Actions script is enough?

blink/memorymalloc.c
trungnt2910: Somehow, this works! Included in the workaround.

trungnt2910:
image

The hello world Linux ELF now works, as well as a naively crafted version of `uname`.

.github/workflows/emscripten.yml
jart: It only needs to be copied into the output folder if you're doing something like minifying it. I don't think we're ready for a JavaScript build system quite yet. Simply copying the non-minified source code into the repo should be sufficient.

jart: Please be to make sure the license and copyright are at the top of the file, if it's not there already.

trungnt2910: Done.

blink/blink.c
Vogtinator: You'd still need to use `-s EXIT_RUNTIME=0` if you want to do any calls after returning from `main`. It's possible that it'll appear to work, but with `-s ASSERTIONS=1` you'll see assert failures.

IMO ideally the API is adapted a bit so that `main` just initializes the system and a separate `exec` function is called for each program invocation. That way the program doesn't attempt initialization multiple times.

trungnt2910: > It's possible that it'll appear to work, but with -s ASSERTIONS=1 you'll see assert failures.

Funny thing is, I _am_ building with assertions:

https://github.com/trungnt2910/blink/blob/73f85932af750723c572208ad2a2e52faa6e4176/.github/workflows/emscripten.yml#L55

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,67 @@
...
+            --shell-file blink/blink-shell.html \
+            -s INVOKE_RUN=0 \
+            -s INITIAL_MEMORY=1073741824 \
+            -s EXIT_RUNTIME=1 \
+            -s EXPORTED_RUNTIME_METHODS=callMain \
+            -fno-builtin-exit \
+            -sASSERTIONS
+
+      - name: Deploy sample to GitHub pages
+        uses: JamesIves/github-pages-deploy-action@v4.4.1
```
Vogtinator, line 58: For 3rd party actions I usually recommend to use the full commit hash to prevent that the tag is retroactively changed. Especially here there is write access to the repo possible.

trungnt2910: Done.

blink/blink.c
Vogtinator: Hm. I can't test it here right now because with `-sASSERTIONS` it complains about a stack overflow and I'm also seeing other weird issues like static variables not getting initialized :-/ Probably caused by an emscripten or llvm update.

trungnt2910: I removed `-s EXIT_RUNTIME=1`. This should revert to the default `-s EXIT_RUNTIME=0`.
In this case, `EXIT_RUNTIME` doesn't even matter because... we never exit the runtime.

The `main` function in `blink.c` never returns, sane Linux binaries always call an `exit` syscall at its termination, which will be translated to a `exit` library call.

And our `exit` library call for Emscripten is already overridden.

trungnt2910: image

Still building with `ASSERTIONS`, no error running two `bash` commands.

Vogtinator: > The main function in blink.c never returns, sane Linux binaries always call an exit syscall at its termination, which will be translated to a exit library call.

Ok. In that case my idea of having `main` separate from an `exec` function woudn't work anyway.

jart: You two did not just get bash running in the web browser!!! I can't believe what I'm seeing. Everyone is going to want this. cc: @ajbouh

jart (approved):
Looks good to me. Thank you, both of you. We've achieved something here that has the potential to become very impactful.

blink/blink.c
trungnt2910: It's actually not such a great deal, only --help and --version works. It's a bit more work if you want to run the real thing, starting from adding the missing faccessat2 syscall.

0xhiro:
Wow! This is impressive! Well done!

blink/blink.c
jart: I'm currently working on `set_robust_list`. I should be able to make `faccessat2` happen within the next 24 hours. Speaking of which, it looks like our `faccessat` implementation is currently broken, so I'll fix that as well.

Aside from those things, do you anticipate any other blockers, aside from figuring out a way to have that setcontext() i/o behavior upon read() that we discussed in the WASM issue earlier?

trungnt2910: 1. The shell included in this pull request is a minimal shell, the default one provided by emscripten. Therefore there currently is no way to run interactive mode on the browser.
2. While the `gh-pages` branch is deployed, you haven't enabled GitHub pages yet, so I can't really see what's happening on your `master` branch.

jart: 1. Implementing support for ANSI codes is relatively straightforward. See [blink/pty.c](blink/pty.c).
2. I've gone into settings and configured Pages to use GitHub Actions. Is there any further action required on my part to enable this?

image

trungnt2910: > I've gone into settings and configured Pages to use GitHub Actions. Is there any further action required on my part to enable this?

Currently, I'm using a custom build action (as I don't think GitHub Pages have built-in support for websites other than popular frameworks) that deploys the compiled pages to a separate branch (in this case it is `gh-pages`).

So your source has to be the `gh-pages` branch instead of GitHub Actions.

trungnt2910: image

This is how I configure my own fork.

jart: https://jart.github.io/blink/blink.html

trungnt2910: On my personal fork, I've re-written the terminal frontend to use `xterm.js`:

image

The first line is `bash` running without arguments (fails with the missing `faccessat2` mentioned above).
Second and third invocations are `--version` and `--help` respectively.

On my fork, the terminal frontend should (theoretically) be able to handle user input.

trungnt2910: @jart About running `bash`:

image

`bash` triggers an assert in the Emscripten library by calling `faccessat2` with `AT_EACCESS`. As `uid` and `gid` does not exist on a browser environment, `blink` should either emulate it (hard) or simply stub `AT_EACCESS` away for browser environments.

Vogtinator: IMO that could be ignored in upstream emscripten. Until that's fixed, blink can just mask `AT_EACCESS` away.

jart: I've masked `AT_EACCESS` away in https://github.com/jart/blink/commit/2cb026d1441c611abbd92d33429b4d1c0f5ac9a0 The big blocker for bash is going to be this:

```
I2023-01-15T14:09:56.445840:blink/syscall.c:306:87720 missing mandatory clone() thread flags: 0x10f00 out of 0x1200000
```

Emscripten doesn't support fork() and Blink is currently implemented to use fork(), rather than simulating it using threads. Even on Linux it currently isn't possible to do the special kind of forking that bash is requesting. Those hex numbers means it wants the kernel to set and clear the child thread id. But across multiple processes, only the kernel is able to do that kind of shared memory synchronization.

jart: One shell that is working really well, is Cosmopolitan Libc's fork of the Almquist Shell, which it calls The Unbourne Shell. Noting that's only on POSIX platforms, where fork() is possible.

```
$ o//blink/blink ~/cosmo/o//examples/unbourne.com
>: ls
HTAGS  LICENSE  Makefile  README.md  TAGS  blink  build  o  test  third_party  tool
>:
```

The unbourne shell is really nice because it has a [bestline](https://github.com/jart/bestline) repl which is just as good as GNU Readline in terms of quality. It's a much surer thing that it'll be compatible with Blink, than a Glibc build of Bash.

PR #30, opened 2023-01-15
Fix bugs in LDS and LES instructions

tkchia:
- get size of offset operand from CPU mode and instruction prefixes (`lds (...),%ax` and `lds (...),%eax` both possible)
- segment value comes after offset value in memory

### DIFF
--- a/blink/machine.c
+++ b/blink/machine.c
@@ -1186,9 +1186,31 @@ static void Op1b8(P) {
 }
 
 static relegated void LoadFarPointer(P, u64 *seg) {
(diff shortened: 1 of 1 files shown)


tkchia:
I am trying to see how far I can go in getting a FreeDOS 1.3 Floppy Edition (http://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/distributions/1.3/official/) boot floppy image to run.

jart (approved):
This is great. Thank you!

PR #32, opened 2023-01-16
Some support for running interactive programs with emscripten

Vogtinator:
Based on @trungnt2910's integration on xterm.js, utilize emscripten's `ASYNCIFY` to be able to read `stdin` during emulation.

Some other changes were necessary to get `busybox` and `python.com` to work, see the individual commits for details.

Screenshot_20230116_204040


blink/blink-shell.html
```diff
@@ -28,6 +29,8 @@
     </style>
     <!-- BLINK SPECIFIC: Enable SharedArrayBuffer -->
     <script src="coi-serviceworker.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/xterm@5.1.0/lib/xterm.min.js"></script>
```
jart, line 32: Could you please download this JS blob into third_party/ and put the contents of https://github.com/xtermjs/xterm.js/blob/master/LICENSE in a comment at the top? Please also put the URL from which it was obtained into the comment. The same goes for the .css file.

blink/syscall.c
```diff
@@ -1501,6 +1538,12 @@ static int IoctlTcsets(struct Machine *m, int fd, int request, i64 addr,
 static int IoctlTiocgpgrp(struct Machine *m, int fd, i64 addr) {
   int rc;
   u8 *pgrp;
+
+#ifdef __EMSCRIPTEN__
+  // Force shells to disable job control in emscripten
+  return -1;
```
jart, line 1544: Be sure to set `errno`. The documented errnos for `tcgetpgrp` are `ENOTTY` and `EBADF`. I'd recommend using `ENOTTY`.

jart (changes requested):

blink/syscall.c
Vogtinator: Done

blink/blink-shell.html
Vogtinator: Done. I'm not sure whether that's technically enough though, because that file itself is built from (and thus contains) multiple other JS sources... With the way the node ecosystem works we'd probably end up with ~1000 license files.

jart: I think this change is a step in the right direction to ensuring we respect the license terms of open source authors. We would ideally vendor the transitive closure of dependencies and setup a build system to compile all this typescript code, however I didn't want to bog you down in process. Contributions are welcome in helping with npm dependency analysis to ensure everyone who contributed is represented. What's most important is that we're minimizing the number of third party web servers our app needs to contact, thereby offering better security and privacy to the users of our code.

jart (approved):
Thank you so much. I can't wait to see it go live!

PR #41, opened 2023-01-20
Haiku build fixes

trungnt2910:
- Add -fPIC for Haiku, somehow GCC requires it for -fPIE to work.
- Fix `speed_t` overflow on Haiku. POSIX does not specify what `speed_t` is, on Linux it is an `int`, but for other platforms it might be something else, even a `struct`. On Haiku, it is an `unsigned char`, making a Haiku `speed_t` variable overflow when storing a Linux baud value.


jart (approved):
Thank you!

PR #42, opened 2023-01-21
Haiku syscall fixes

trungnt2910:
- SOL_SOCKET is defined as -1 on Haiku, therefore XlatSocketLevel has to be changed.
- Haiku _recognizes_ PE binaries, but cannot `exec` them. `execve` on Haiku therefore fails but does not return `ENOEXEC` for Cosmopolitan binaries. Adding an additional error code check fixes this.

This is the last fix required for all blink `make check` to pass on Hyclone 😄

### DIFF
--- a/blink/syscall.c
+++ b/blink/syscall.c
@@ -53,6 +53,7 @@
 
 #ifdef __HAIKU__
 #include <sys/sockio.h>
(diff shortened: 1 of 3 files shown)


jart (approved):
Thank you!

PR #43, opened 2023-01-21
blink-shell.html: Set exec permissions on the executable file

Vogtinator:
Currently it fails with `command not found: executable`


jart (approved):
Thank you!

PR #44, opened 2023-01-21
Fix read syscall errno

trungnt2910:
This fixes one more test on real Haiku.


jart (commented):
But it's listed by POSIX.1 as a valid error. So if the cosmo test assuming `EAGAIN` then do you think it should be updated? https://pubs.opengroup.org/onlinepubs/009604599/functions/read.html

trungnt2910:
Yeah, after re-reading the man pages, I think this issue should be closed and the Cosmo test should be updated instead.

>        Other errors may occur, depending on the object connected to fd.

PR #45, opened 2023-01-21
Fix GDT limit check when loading selector into segment register

tkchia:
This issue was causing Amey Narkhede's "Game of Life" bare metal image (https://github.com/glitzflitz/gameoflife) to crash under Blinkenlights shortly after startup.

### DIFF
--- a/blink/machine.c
+++ b/blink/machine.c
@@ -129,9 +129,9 @@ static relegated void OpMovEvqpSw(P) {
 
 static relegated int GetDescriptor(struct Machine *m, int selector,
                                    u64 *out_descriptor) {
-  unassert(m->system->gdt_base + m->system->gdt_limit <= kRealSize);
+  unassert(m->system->gdt_base + m->system->gdt_limit < kRealSize);
   selector &= -8;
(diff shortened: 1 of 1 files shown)


jart (approved):
Thanks!

jart:
I'd also like to extend you a formal invitation to our Discord https://discord.gg/nMvdGvyz

tkchia:
Hello @jart,

Thank you!  Let me see (again) if I can join.

There are a few more subtleties surrounding the descriptor tables &mdash; e.g. the length of the linear address in `lgdt` under various conditions &mdash; which I also hope to fix.  I am not sure how to properly do regression testing for these though.

Thank you!

jart:
What we could probably do is set up a folder called `test/metal/` which follows the same basic principle as `test/func/` where it maybe assembles some real mode code and then runs it in Blink using the `-r` flag. The program could indicate whether it's passed or fail, possibly using the `syscall` instruction to call exit.

tkchia:
Hello @jart,

Let me see what I can do towards this.  Thank you!

PR #46, opened 2023-01-21
Execve prioritizes emulation over native execution

trungnt2910:
Now Execve tries to execute the binary with blink first, then falls back to using the host's native `execve`.

This behavior is useful even on native Linux, as we don't really expect binaries to be able to escape from the blink sandbox.

(Imagine some kind of Linux malware checking for `GenuineBlink` before `execve`ing itself...)


jart (approved):
Blink is still unsecured from a sandboxing perspective, but I'm glad you're keeping that use case in mind. We'll need to explain to anyone who tries to run `blink gcc` why it goes slower than `qemu-x86_64 gcc` since qemu takes the optimistic approach of calling execve() first. However I feel like this is the right thing to do, since execve() presents other problems like binfmt_misc, to which qemu actually registers itself. So I don't want people to ask for Blink and then get Qemu.

PR #48, opened 2023-01-25
Blinkenlights: emulate BIOS int 0x13 extended disk read (ah = 0x42)

tkchia:
Used by Narkhede's "Game of Life" PC booter image (https://github.com/glitzflitz/gameoflife).

### DIFF
--- a/blink/blinkenlights.c
+++ b/blink/blinkenlights.c
@@ -2584,6 +2584,66 @@ static void OnDiskServiceReadSectors(void) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


tkchia:
With this patch, the "Game of Life" boot sector can now load and fire up the second stage bootloader:

image

(The binary later crashes though when it does a `lretl` to transition to 32-bit protected mode.  This will probably need to be fixed separately.)

image

image

Thank you!

jart (approved):
Awesome. Thank you!

PR #49, opened 2023-01-27
Partly implement descriptor caches for segment registers

tkchia:
This patch changes the internal implementation of the emulated segment registers, so that each segment register comprises

- a visible portion, containing the selector value (protected/ long mode) or paragraph value (real mode), which is directly accessible via `mov`, `push`, etc.
- a hidden portion, containing the segment base address.

This makes the implementation closer to what is described in the Intel SDM Volume 3A, §3.4.3 (Segment Registers).

This patch allows the third stage bootloader of "Game of Life" (https://github.com/glitzflitz/gameoflife) to run.

The Blinkenlights register display formatting is also changed: if any segment register bases are non-zero, Blinkenlights displays the visible segment value _and_ the segment base for each segment register, e.g.
```
FS  0000 (@00000000)  DS  1fe0 (@0001fe00)  CS  1fe0 (@0001fe00)
GS  0000 (@00000000)  ES  1fe0 (@0001fe00)  SS  1fe0 (@0001fe00)
```

### DIFF
--- a/blink/machine.h
+++ b/blink/machine.h
@@ -72,6 +72,18 @@
 #define kBusCount   256       // # load balanced semaphores in virtual bus
 #define kBusRegion  kSemSize  // 16 is sufficient for 8-byte loads/stores
 
+#define CR0_PE 0x01       // protected mode enabled
+#define CR0_MP 0x02       // monitor coprocessor
+#define CR0_EM 0x04       // no x87 fpu present if set
+#define CR0_TS 0x08       // task switched x87
+#define CR0_ET 0x10       // extension type 287 or 387
+#define CR0_NE 0x20       // enable x87 error reporting
+#define CR0_WP 0x00010000 // write protect read-only pages @pl0
+#define CR0_AM 0x00040000 // alignment mask
+#define CR0_NW 0x20000000 // global write-through cache disable
+#define CR0_CD 0x40000000 // global cache disable
+#define CR0_PG 0x80000000 // paging enabled
+
 #define PAGE_V    0x0001  // valid
 #define PAGE_RW   0x0002  // writeable
 #define PAGE_U    0x0004  // permit user-mode access
--- a/blink/machine.h
+++ b/blink/machine.h
@@ -161,14 +161,23 @@ struct MachineMemstat {
   memstat_t pagetables;
 };
 
+// Segment descriptor cache
+// @see Intel Manual V3A §3.4.3
+// Currently only the base address for each segment register is maintained;
+// a full implementation will also cache its limit & access rights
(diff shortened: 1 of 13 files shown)


tkchia:
I realize that this PR is rather intrusive; let me know if it may need a redo.  Thank you!

ghaerr:
Hello @tkchia,

Wow! Very nice. Does this work then complete the implementation of the x86 "unreal" mode, where real mode execution of instructions with 32-bit addresses (i.e. 16-bit instructions with addr32 prefix) will use the previously set via protected-mode GDT hidden base address of the referenced segment register for the address calculation?

Thank you!

tkchia:
@ghaerr : I would say that this patch is more about filling in the implementation of good old _protected_ mode.

blink/machine.h
```diff
@@ -161,14 +161,23 @@ struct MachineMemstat {
   memstat_t pagetables;
 };
 
+// Segment descriptor cache
+// @see Intel Manual V3A §3.4.3
+// Currently only the base address for each segment register is maintained;
+// a full implementation will also cache its limit & access rights
+struct DescriptorCache {
+  u16 sel;   // visible selector value (or paragraph value in real mode)
+  u64 base;  // base linear address
```
jart, line 170: Nit but there's six bytes of wasted space here. You could create two separate arrays.

blink/memory.c
```diff
@@ -112,7 +112,8 @@ static u64 FindPageTableEntry(struct Machine *m, u64 page) {
 u8 *LookupAddress(struct Machine *m, i64 virt) {
   u8 *host;
   u64 entry, page;
-  if (m->mode != XED_MODE_REAL) {
+  if (m->mode == XED_MODE_LONG ||
+      (m->mode != XED_MODE_REAL && (m->system->cr0 & 0x80000000) != 0)) {
```
jart, line 116: It'd be nice to have a constant for this number.

jart (commented):
I like this change. It's good to have better support for good old protected mode. This change might also make it possible for us to support an esoteric feature from the Linux ABI where you read directly from the GDT https://github.com/jart/cosmopolitan/blob/f422edaf15f0fb358632b8741466e58c70a07ced/libc/nexgen32e/rdtscp.h#L36-L61

blink/memory.c
tkchia: @jart : perhaps I can bring in some constant names from the Cosmopolitan project.  Thank you!

blink/machine.h
jart: It's not worth thinking that much about.

jart (approved):
Thank you!

PR #51, opened 2023-02-01
Tweak emulation of CPUID leaf 0x80000000, and add WBINVD instruction

tkchia:


blink/cpuid.c
```diff
@@ -64,8 +64,11 @@ void OpCpuid(P) {
   ax = bx = cx = dx = 0;
   switch (Get32(m->ax)) {
     case 0:
-    case 0x80000000:
       ax = 7;
+      goto vendor;
+    case 0x80000000:
+      ax = 0x80000007;
```
jart, line 70: This means we *should* implement CPUID leaves

- 80000001H
- 80000002H
- 80000003H
- 80000004H
- 80000005H
- 80000006H
- 80000007H

too, no?

tkchia: @jart : indeed, you are right.  Perhaps I should limit the return maximum leaf value to `0x80000001` for the time being (since `0x80000001` is implemented).  I will make the change.  Thank you!

jart (approved):

PR #55, opened 2023-02-04
Fix Cygwin JIT memory protection after fork

### LINKED ISSUE #27: JIT on Windows
Opening this issue to track the problems preventing JIT on Windows from happening yet.

From my own experience, the problem is due to the difference between Windows calling convention and Sys-V.

- [x86_64 Windows](https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170) and x86_64 Sys-V passes arguments in different registers. A solution might be just keep using the Sys-V ABI in the JITted code and make a trampoline before returning to compiled Windows code.
- FS - GS register differences. Although as JIT already works on macOS (which also uses different FS/GS than Linux), I guess the same can be applied to Windows?
- x86_64 Sys-V requires a 128-byte red zone below `rsp`, while x86_64 Windows does not. When jumping to JIT code, the value of `rsp` must be 128 bytes lower than it is supposed to be, and all read/write operations involving `rsp` must be hooked to modify the value. Things will get tricky when dealing with opcodes such as `call`,...

trungnt2910:
Closes #27 

image

`glxgears` look quite fast.

I decided to put the fix protection to `jit.c` as it looks more like a general JIT stuff and not specific to Cygwin. Hopefully, compilers may be able to optimize this out on platforms that don't use the workaround.

### DIFF
--- a/blink/jit.c
+++ b/blink/jit.c
@@ -386,6 +386,22 @@ int DisableJit(struct Jit *jit) {
   return 0;
 }
 
+/**
+ * Fixes the memory protection for existing Just-In-Time code blocks.
+ */
+int FixJitProtection(struct Jit *jit) {
+  struct Dll *e;
+  LOCK(&jit->lock);
+  int prot;
+  prot = atomic_load_explicit(&g_jit.prot, memory_order_relaxed);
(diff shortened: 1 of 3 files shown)


trungnt2910:
cc @ClickNin

jart (approved):
Outstanding. Thank you.

PR #58, opened 2023-02-06
Fix utf-8 printing on windows

JoshuaWierenga:
Includes the SetConsoleOutputCP call we spoke about with the only change being that I found a constant for the code page so it's not a random integer now. Currently, the ifdef specifically checks for msvc, mingw-w64 or cygwin in non posix modes since I am unsure if the issue also occurs on cygwin in posix mode.

Is it also worth including a temporary fix for sizeof(wchar_t) = 2 on windows since that is an issue in at least pty.c?

### DIFF
--- a/blink/blink.c
+++ b/blink/blink.c
@@ -243,6 +243,12 @@ void exit(int status) {
 
 int main(int argc, char *argv[]) {
   SetupWeb();
+  // Ensure utf-8 is printed correctly on windows, this method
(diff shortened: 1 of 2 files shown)


jart (approved):
Thank you!

jart:
> Is it also worth including a temporary fix for sizeof(wchar_t) = 2 on windows since that is an issue in at least pty.c?

If you need it, we can fix it.

PR #65, opened 2023-02-15
Ensure configure tests run when using mingw-w64

JoshuaWierenga:
Originally all tests other than flock failed when using mingw-w64 as cc since it sticks `.exe` on the end of all executable names which confuses the configure script into thinking the features test binaries don't exist.

There is also a fix included for random tests not working with dev_urandom checking the host's /dev/random which exists if building on wsl and I assume cygwin despite this obviously not being the case if the resulting blink binary is run from windows directly or on msys. And rtlgenrandom using a header that doesn't exist in mingw-w64's header library, despite this, the function is available. The rtlgenrandom check could be simpler if I knew that cygwin only defined _WIN32 for native building. Otherwise I think every check has to become the mingw-w64 specific check which needs one of a few system headers to be included first.

I have tried enabling pthreads with {i686, x86_64}-w64-mingw32-gcc-10-posix but just get a windows error popup saying libwinpthread-1.dll was not found so need to investigate that.

Additionally, repeated calls to ./configure can cause some tests, mostly memccpy and __int128 to error with either "Text file busy" and at least one error but I can't reproduce it currently.

Edit: Ignore my multiple title changes, I have been up too many hours and apparently can't read my own title correctly.

### DIFF
--- a/configure
+++ b/configure
@@ -184,6 +184,12 @@ compile() {
 
 config() {
   PROGRAM="$1"
+  # If running on cygwin, msys2 or wsl then enable testing mingw-w64 produced binaries
+  if test "${CC#*w64-mingw}" != "${CC}"; then
+    RUNPROGRAM="$1.exe"
+  else
+    RUNPROGRAM="$1"
+  fi
   MESSAGE="$2"
   shift 2
   LOG=$(mktemp blink-config.log.XXXXXX)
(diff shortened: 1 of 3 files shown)


JoshuaWierenga:
Whatever macro to check for native windows and not cygwin it decided on also has to be added to random.c.

jart (approved):
Thank you!

> I have tried enabling pthreads with {i686, x86_64}-w64-mingw32-gcc-10-posix but just get a windows error popup saying libwinpthread-1.dll was not found so need to investigate that.

Yeah I forgot to do something to address that. I'll fix it right now.

Issue #53, opened 2023-02-03
Some (remaining) issues in getting "Game of Life" to work under Blinkenlights


tkchia:
There are some issues in Blinkenlights currently preventing Amey Narkhede's "Game of Life" (https://github.com/glitzflitz/gameoflife) PC booter program from working (`blinkenlights -r gameoflife.bin`).  These issues may need quite a fair amount of work to fully resolve.

* The "Game of Life" program uses "huge pages" (2 MiB), which are supported by the x86-64 CPU, but not yet emulated by Blinkenlights.
* (fixed by https://github.com/jart/blink/commit/5121185f92628f2d27c71f85a7df50ed18b8263a) Before using "huge pages" &mdash; but after switching to 64-bit long mode &mdash; the program does a `mov $0,%bx; mov %bx,%ss` for some reason.  This causes Blinkenlights to crash the program.
  (QEMU allows the program to continue.  ~~But it is possible that QEMU is wrong here.  We probably need to check what an actual x86-64 CPU does.~~)
  ![image](https://user-images.githubusercontent.com/5507843/216626236-4d03567a-59a9-49f9-bc1a-c30e18bcf2d2.png)

Thank you!

tkchia:
> (QEMU allows the program to continue. But it is possible that QEMU is wrong here. We probably need to check what an actual x86-64 CPU does.)

It turns out that QEMU's emulation (see [`target/i386/seg_helper.c`](https://gitlab.com/qemu-project/qemu/-/blob/b585edca34a817fdb751dfe94fbd3cde32ffe60d/target/i386/tcg/seg_helper.c#L1278)) is apparently correct, and as far as I can tell, matches the behavior seen on a real CPU (!).

In particular, it is OK to load a zero selector value into `%ss` in long mode &mdash; while in ring 0.  In ring 3 though, loading zero into `%ss` will raise a CPU exception.

Thank you!

PR #69, opened 2023-02-18
Tweaks for segment selector handling and page table handling

tkchia:
* Allow loading null selector into segment register in some cases (partly addresses https://github.com/jart/blink/issues/53)
* Do not require `PAGE_V` to be set on `cr3` &mdash; when reading top-level page table
* Do not require `PAGE_U` to be set on page table entries, when guest is not running at Ring 3
* Initialize `cs` and `ss` to Ring 3 selector values, when guest is starting in non-metal mode


jart (approved):
Awesome change, and thanks for the tests!

The test you sent me in your last change actually saved me from making a regression. I was able to quickly address what broke before I pushed it to GitHub. So thank you for that.

tkchia:
Thanks @jart !  I am glad my earlier work was of help. :slightly_smiling_face:

PR #71, opened 2023-02-20
Implement huge virtual memory pages, and fix some minor issues

tkchia:
* Fix disassembly of `set`_cc_ instructions &mdash; e.g. `0f 90 c1` should be `seto %cl`
* Reset entire TLB on each `%cr3` or `%cr4` load
  * Intel manual (V3A §2.5) suggests that TLB is flushed on at least _some_ loads into `%cr3` or `%cr4`
* Implement huge (1 GiB / 2 MiB) virtual memory pages (https://github.com/jart/blink/issues/53)

### DIFF
--- a/blink/memory.c
+++ b/blink/memory.c
@@ -120,6 +120,21 @@ u64 FindPageTableEntry(struct Machine *m, u64 page) {
     index = (page >> level) & 511;
     entry = Get64(GetPageAddress(m->system, table, level == 39) + index * 8);
     if (!(entry & PAGE_V)) return 0;
+    if (m->metal) {
+      entry &= ~(u64)(PAGE_RSRV | PAGE_HOST | PAGE_MAP |
+                      PAGE_EOF | PAGE_MUG | PAGE_FILE);
+    }
+    if ((entry & PAGE_PS) && level > 12) {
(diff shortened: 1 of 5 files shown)


tkchia:
Screenshot from 2023-02-21 13-54-54

blink/memory.c
```diff
@@ -120,6 +120,21 @@ u64 FindPageTableEntry(struct Machine *m, u64 page) {
     index = (page >> level) & 511;
     entry = Get64(GetPageAddress(m->system, table, level == 39) + index * 8);
     if (!(entry & PAGE_V)) return 0;
+    if (m->metal) {
+      entry &= ~(u64)(PAGE_RSRV | PAGE_HOST | PAGE_MAP |
+                      PAGE_EOF | PAGE_MUG | PAGE_FILE);
+    }
+    if ((entry & PAGE_PS) && level > 12) {
+      // huge (1 GiB or 2 MiB) page; "rewrite" the TLB copy of the page table
```
jart, line 128: It's really nice to see this change, because this could potentially be used to speed up our `mmap()` implementation in userspace. One of the issues is if I mmap() a petabyte of memory, it needs to create thousands of page tables. I can think of a few ideas we could create higher level page table entries that represent all 512 entries in the table beneath.

jart (approved):
Thank you!

PR #76, opened 2023-03-06
VFS feature

trungnt2910:
This PR contains the "VFS feature" as discussed in the Discord server.

As this feature is expensive in terms of syscalls and memory, it is disabled by default.
Before this PR I've made a reasonable attempt to fix as many memory leaks as possible by running the tests with `MODE=asan`.

I've also left some more details in the commit messages.


trungnt2910:
Did this PR cause any build warning/failures? I've tested both on Cygwin and Linux and ensured that there are no build failures and/or warnings.

blink/vfs.h
```diff
@@ -0,0 +1,428 @@
...
+#define VfsTcdrain      tcdrain
+#define VfsSockatmark sockatmark
+#define VfsExecve       execve
+#define VfsMmap         mmap
+#define VfsMunmap       munmap
+#define VfsMprotect     mprotect
+#define VfsMsync        msync
+#endif
+
+#define VFS_SYSTEM_ROOT_MOUNT "/SystemRoot"
```
jart, line 426: Can we not hard code this? Shouldn't this be behind the `-C` flag?

jart (changes requested):
Please apply the changes from 6d72c93bfd8816f7c8d71e795f4141bffb4066e7 you may want to consider adding your username to `build/config.mk` so that `-Werror` is enabled in your dev environment. I don't enable it by default for obvious reasons, but it should be enabled for development.

blink/vfs.h
trungnt2910: > I don't enable it by default for obvious reasons, but it should be enabled for development.
I see that enabling this breaks the build, even after cherry-picking 6d72c93:

```
blink/disinst.c: In function ‘DisName’:
blink/disinst.c:121:37: error: array subscript has type ‘char’ [-Werror=char-subscripts]
  121 |     for (np = name; *np && (islower(*np) || isdigit(*np)); ++np) {
      |                                     ^~~
cc -fno-align-functions -fno-common -pthread -fpie -g -O2 -fno-omit-frame-pointer -fno-optimize-sibling-calls -gdwarf-2 -U_FORTIFY_SOURCE -Wall -Werror -Wno-unused-function -isystem third_party/libz -D_FILE_OFFSET_BITS=64 -D_DARWIN_C_SOURCE -D_DEFAULT_SOURCE -D_BSD_SOURCE -D_GNU_SOURCE -iquote.  -c -o o//blink/divmul.o blink/divmul.c
blink/disinst.c:121:53: error: array subscript has type ‘char’ [-Werror=char-subscripts]
  121 |     for (np = name; *np && (islower(*np) || isdigit(*np)); ++np) {
      |                                                     ^~~
```

Should I fix this?

trungnt2910: The `-C` flag is the equivalent of the `$BLINK_PREFIX` env var in this PR.
I'll edit it to respect `-C` as well.

`/SystemRoot` is a default `hostfs` mount, similar to the default `devfs` mount at `/dev` and `procfs` mount at `/proc`. This should suffice until we implement `fstab` support.

trungnt2910: I've fixed my code to respect the `-C` flag when mounting `/` as `hostfs`.
The system root `/` is still mounted to `/SystemRoot`, unless you think that I should keep that under a flag too.

jart (approved):

PR #81, opened 2023-03-12
VFS: CI and a few small bug fixes

trungnt2910:
- Added CI builds with `--enable-vfs`.
- Fixed bug when reading `/SystemRoot` on Cygwin.
- Fixed some memory leaks when handling UNIX sockets.
- Added fallback to `execve` on systems without `fexecve` in `HostfsFexecve`.


jart (approved):
Ohh so that's why `munmap_test` wasn't working. I doubt there's much we can do about that other than disabling it.

Thanks for the requested changes!

PR #88, opened 2023-03-16
VFS: Initial procfs support

trungnt2910:
This patch contains changes for procfs support, as well as fixing some bugs in the VFS system:
- Fixed some more buffer overflow bugs.
- Fixed traversing across mount boundaries.
- Fixed `errno` when doing an unsupported operation on a device. (`EPERM`, not `EOPNOTSUPP`).

Currently only a few static files (`filesystems`, `meminfo`, `mounts`, `self`, `uptime`), as well as the folder corresponding to the current process and its static contents (`cwd`, `exe`, `mounts`, `root`) are supported. For further support, `blink` must have some method to register, manage and query shared process information.

### DIFF
--- a/blink/hostfs.c
+++ b/blink/hostfs.c
@@ -128,6 +128,21 @@ cleananddie:
   return -1;
 }
 
+int HostfsReadmountentry(struct VfsDevice *device, char **spec, char **type, char **mntops) {
+  struct HostfsDevice *hostfsdevice = (struct HostfsDevice *)device->data;
+  *spec = strdup(hostfsdevice->source);
+  if (*spec == NULL) {
+    return enomem();
+  }
+  *type = strdup("hostfs");
+  if (*type == NULL) {
+    free(*spec);
+    return enomem();
+  }
+  *mntops = NULL;
+  return 0;
+}
+
 int HostfsFreeInfo(void *info) {
   struct HostfsInfo *hostfsinfo = (struct HostfsInfo *)info;
   if (info == NULL) {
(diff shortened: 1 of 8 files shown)


blink/hostfs.c
```diff
@@ -128,6 +128,21 @@ int HostfsInit(const char *source, u64 flags, const void *data,
   return -1;
 }
 
+int HostfsReadmountentry(struct VfsDevice *device, char **spec, char **type, char **mntops) {
+  struct HostfsDevice *hostfsdevice = (struct HostfsDevice *)device->data;
+  *spec = strdup(hostfsdevice->source);
+  if (*spec == NULL) {
+    return enomem();
```
jart, line 135: You can just return -1 in cases like this, because strdup already set the errno.

blink/hostfs.c
```diff
@@ -128,6 +128,21 @@ int HostfsInit(const char *source, u64 flags, const void *data,
   return -1;
 }
 
+int HostfsReadmountentry(struct VfsDevice *device, char **spec, char **type, char **mntops) {
+  struct HostfsDevice *hostfsdevice = (struct HostfsDevice *)device->data;
+  *spec = strdup(hostfsdevice->source);
```
jart, line 133: One best practice for API design in C is that the output parameters should only be modified on success. There's very few exceptions to this rule. But it's mostly an issue for public APIs.

blink/procfs.c
```diff
@@ -16,94 +16,1463 @@
...
+  *info = malloc(sizeof(struct ProcfsInfo));
+  if (*info == NULL) {
+    return enomem();
+  }
+  if (parent == NULL) {
+    **info = g_piddirinfos[0];
+    // TODO(trungnt): This should be dynamically allocated when we support
+    // more processes.
+    (*info)->ino = PROCFS_FIRST_PID_INO;
+    sprintf((*info)->name, "%d", pid);
```
jart, line 210: I recommend against using sprintf() because it's one of those functions that triggers security people and attracts attention from scanners.

jart (approved):
Thank you!

blink/procfs.c
trungnt2910: I used sprintf at that once instance as I wonder how a 32-bit int can take more than   16 chars to represent.

PR #91, opened 2023-03-17
Fix issues in fast versions of rep movsb & rep stosb

tkchia:
The `diactual` & `siactual` pointers are now offset by the `ES` & `DS` bases after the first iteration of the outer loop.  This is needed for most real-mode uses of `rep movsb` to work.

I also try to handle the (hopefully rare) case of `DI` or `SI` "wrapping around" to 0 or -1 in the middle of a `rep movsb` or `rep stosb`.

### DIFF
--- a/blink/string.c
+++ b/blink/string.c
@@ -68,40 +68,37 @@ static void WriteInt(u8 p[8], u64 x, unsigned long w) {
   }
 }
 
-static u64 AddDi(P, u64 x) {
+static void AddDi(P, u64 x) {
   u64 res;
   switch (Eamode(rde)) {
     case XED_MODE_LONG:
-      Put64(m->di, (res = Get64(m->di) + x));
+      Put64(m->di, Get64(m->di) + x);
       break;
     case XED_MODE_LEGACY:
-      Put64(m->di, (res = (Get32(m->di) + x) & 0xffffffff));
+      Put64(m->di, (Get32(m->di) + x) & 0xffffffff);
       break;
     case XED_MODE_REAL:
-      Put16(m->di, (res = Get16(m->di) + x));
+      Put16(m->di, Get16(m->di) + x);
       break;
     default:
       __builtin_unreachable();
   }
-  return res;
 }
 
 static u64 AddSi(P, u64 x) {
-  u64 res;
   switch (Eamode(rde)) {
     case XED_MODE_LONG:
-      Put64(m->si, (res = Get64(m->si) + x));
+      Put64(m->si, Get64(m->si) + x);
       break;
     case XED_MODE_LEGACY:
-      Put64(m->si, ((res = Get32(m->si) + x) & 0xffffffff));
+      Put64(m->si, (Get32(m->si) + x) & 0xffffffff);
       break;
     case XED_MODE_REAL:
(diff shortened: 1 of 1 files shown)


tkchia:
The segment base offset issue cropped up when running the initial `rep movsb` of the FreeDOS 1.0 kernel (in the `fdboot.img` image at https://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/distributions/1.0/).

ghaerr:
Hello @tkchia,

> the FreeDOS 1.0 kernel (in the fdboot.img image at https://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/distributions/1.0/).

I tried running this image in `blink16` just for fun and it halted after encountering an 0F prefix (non-8086) opcode midway during the boot. Are the standard FreeDOS images all built for 80386 now? I'm finding that a lot of DOS programs, including some in the FreeDOS distribution, assume a 386+ CPU without testing for it. This means that perhaps blink16's DOS syscall implementation should be added to blink, should there be any interest.

Thank you!

tkchia:
Hello @ghaerr,

I am now (also) testing Blinkenlights with the `720k/x86BOOT.img` image inside the the FreeDOS 1.3 floppy edition 720 KiB boot disk.  (See `720k/x86BOOT.img` inside `FD13-FloppyEdition.zip` at https://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/distributions/1.3/official/ .)

The kernel in FreeDOS 1.3 also triggers the same `rep movsb` issue addressed by this PR.

You might also want to try running this image under `blink16`. :slightly_smiling_face: 

Thank you!

ghaerr:
> I am guessing that was a mistake in FreeDOS 1.0's kernel. Not cool.
> Later versions of the FreeDOS kernel have an explicit CPU check after saying 123.

It is failing after displaying 123, but I haven't spent time to determine whether this is before after a CPU type check. (Also, without a symbol table and knowledge of FreeDOS startup, its kind of a pain.)

It's also possible that `blink16` is failing an 8086 CPU check. I have looked at the current FreeDOS version's CPU check and it looks very good, in fact I'm thinking we should use that in ELKS to ward off XMS/386 issues ourselves). I added additional code to the blink16 VM to set the high four bits of FLAGS at startup, but this isn't enforced for all FLAG word overwrites. The FreeDOS CPU check requires this for 8086 determination.

> See 720k/x86BOOT.img inside FD13-FloppyEdition.zip 
> You might also want to try running this image under blink16.

I will try it and report back.

ghaerr:
Hello @tkchia,

FreeDOS v1.3 boots successfully, but later (seemingly during AUTOEXEC.BAT execution?) fails to load strings resources into memory. So not sure whether CPU test was skipped in v1.0, but now boots. Not sure whether strings load failure is a `blink16` emulation issue or a configuration problem. Does `blink` boot FreeDOS now?

Screen Shot 2023-03-22 at 11 15 57 AM

Thank you!

tkchia:
Hello @ghaerr,

> Does `blink` boot FreeDOS now?

Under Blinkenlights &mdash; with this patch &mdash; the FreeDOS kernel halts on an `int $0x12` (which is still unimplemented).

Thank you!

PR #97, opened 2023-04-07
Fix handling of `int` instructions for both metal & non-metal

tkchia:
- `int 0` now behaves like a divide by zero in metal mode, but with `cs:ip` pointing _after_ the `int 0`
- In non-metal (Linux) mode, `int` _n_ with _n_ other than 1, 3, or `0x80`, should raise a `SIGSEGV` (`int 0x80` is unimplemented)
- `int 4` in non-metal mode is a bit of a special case


jart (approved):
LGTM. Thanks!

PR #101, opened 2023-04-14
`iret`, `into`, & far jump/call improvements

tkchia:


jart (approved):
Looks good. Really happy to see `into` and `iret`. Please make the flags magnum more readable per offline discussion. You're encouraged to move the metal tests into `make check` rather than `make emulates` considering their quality.

tkchia:
@jart : thanks!  I will work on these.

PR #102, opened 2023-04-15
Implement interrupt vectoring for real mode

tkchia:
These changes allow a guest program to install their own interrupt handlers in the real mode interrupt vector table (at `0:0`), and also invoke Blinkenlights's default software interrupt implementations.
* The VM now implements two "hypervisor trap" instructions, which I name `hvcall` and `hvjmp`.
* Blinkenlights now also installs a stub BIOS containing default routines for `int 0x10`, `int 0x11`, etc., which call out to the Blink hypervisor.
* Blinkenlights now starts running code from `0xffff:0` rather than `0:0x7c00`.  Most of the legwork for booting up a disk image is now contained in the default implementation for `int 0x19` (which `0xffff:0` calls).


ghaerr:
Hello @tkchia,

Nice work, pretty cool stuff here! A few comments:

From what I understand, the "C" BIOS initially sets up the IVT such that a call to the internal `HaltMachine` entry point is executed, unless the IVT is rewritten by a program, in which case the hardware interrupt handling mechanism is executed. In the case of the default IVT, each vector, rather than pointing to an `hvcall` instruction, points to an array of special  `hvjmp` instructions which are used to `IRET` from the current stack and then execute the internal `HaltMachine` entry point. 

My first comment is that `hvjmp` is a bit confusing name, IMO perhaps something like `hvisr` (for interrupt service request) might be better, as I was initially confused thinking it was a jump instruction similar to call until I read your documentation. I guess the second comment would be that you feel using a new special instruction would be better than pointing the default IVT entries to a section of BIOS code that would fiddle with the stack using normal instructions before executing an `hvcall`? I suppose the problem here may be how to rearrange for the return CS:IP to be properly set before executing the `hvcall`.

For the `hlt` execution case where interrupts are enabled, is `sched_yield` actually releasing the current `blink` timeslice, or is there something else that does? Do we really want to yield `blink` execution, or perhaps actually wait for an interrupt (not yet implemented) or just do nothing? (The ELKS kernel for instance now executes a `hlt` in its idle loop, and that works well for waiting for the hardware INT 8 timer. When `blink` implements a hardware INT 8 timer, likely required for OS emulations, I suppose releasing the current time slice would be a good thing to do).

For the INT 0x18 hook in the case of boot sector read failure - won't that just cause yet another immediate failure, since the IVT is default setup to call `HaltMachine` for it? It might be better to just give an immediate read failure rather than having a secondary failure be displayed in the case the image boot sector can't be read.

I have a couple other small comments I'll post next to the code itself.

I'm thinking of implementing your `hvcall` instruction in `blink16`. Are you thinking that for interrupts >= 0x80, we'll just use the disp16 form of the instruction, at least for the time being?

Thank you!

blink/loader.c
```diff
@@ -471,20 +472,36 @@ static bool LoadElf(struct Machine *m,  //
...
+  memset(m->system->real + 0x00100000, 0, kRealSize - 0x00100000);
   Write16(m->system->real + 0x400, 0x3F8);
   Write16(m->system->real + 0x40E, 0xb0000 >> 4);
+  Write16(m->system->real + 0x410, 1 << 0 |  // floppy drive
+                                   1 << 1 |  // math coprocessor
+                                   3 << 4 |  // initial video mode
+                                   0 << 6 |  // no. of floppy drives - 1
+                                   1 << 9);  // no. of serial devices
   Write16(m->system->real + 0x413, 0xb0000 / 1024);
   Write16(m->system->real + 0x44A, 80);
```
ghaerr, line 495: Perhaps all these BDA (BIOS Data Area) initializations should move to `blink/defbios.c`, to keep BIOS-specific initializations out of the general `blink` INT 19h (or other) interrupt handlers? I like the idea of a separate defbios.c (even if only temporary until the ASM BIOS is implemented), as it keeps real mode BIOS initializations in a single place.

blink/defbios.h
```diff
@@ -0,0 +1,14 @@
...
+
+#define kBiosBase  0x000F0000        // nominal base address of BIOS image
+#define kBiosSeg   (kBiosBase >> 4)  // nominal base segment of BIOS image
+#define kBiosEnd   0x00100000        // address immediately after BIOS image
+#define kBiosEntry 0x000FFFF0        // entry point
+
+void LoadDefaultBios(struct Machine *);
+void SetDefaultBiosIntVectors(struct Machine *);
+
+#endif /* BLINK_TUNABLES_H_ */
```
ghaerr, line 14: BLINK_TUNABLES_H vs BLINK_DEFIOS_H ?

blink/loader.c
tkchia: @ghaerr : I have made the change.  Thanks!

tkchia:
Hello @ghaerr,

> My first comment is that `hvjmp` is a bit confusing name, IMO perhaps something like `hvisr` (for interrupt service request) or or `hvicall` or `hvdefisr` might be better, as I was initially confused thinking it was a jump instruction similar to call until I read your documentation.

Well, my original intention was to convey the idea that it is a "jump-like" instruction which implements a sort of tail call &mdash; by analogy to how a `call` that returns to the next instruction while a `jmp` returns to the caller instead.  Perhaps something like `hvchain` or `hvtailcall` might be better (?), but I am not sure.

@jart : do let me know what you think about this as well.  Thank you!

> I guess the second comment would be that you feel using a new special instruction would be better than pointing the default IVT entries to a section of BIOS code that would fiddle with the stack using normal instructions before executing an hvcall? I suppose the problem here may be how to rearrange for the return CS:IP to be properly set before executing the hvcall.

Yes, this is one problem I had thought a bit about before adopting this (somewhat clunky) arrangement.  For "normal" BIOS services such as `int 0x13` one might be able to say something like
```
isr_0x12:
  hvcall 0x12
  iretw
```
or even
```
isr_0x13:
  sti
  hvcall 0x13
  lretw $2
```
But for something like the `int3` handler, we do actually want the hypervisor to see the `%cs:%rip` at the site of the actual breakpoint, not at the site of the ISR.  This is well nigh impossible to do by fiddling with the stack.
```
isr_0x03:
  /* ??? */
```

A _workable_ alternative might be to define only a single `hvcall` instruction, but designate certain hypervisor call numbers (say, `0x4000`&mdash;`0x40ff`) as returning to an ISR's caller rather than to the instruction after the `hvcall`.  But I felt that such an arrangement would also be quite messy to design and implement.

> It might be better to just give an immediate read failure rather than having a secondary failure be displayed in the case the image boot sector can't be read.

True.  I am not very sure though if it is safe to throw a `HaltMachine(`...`)` from within a routine that is itself called from a `HaltMachine(`...`)` handler...

> Are you thinking that for interrupts >= 0x80, we'll just use the disp16 form of the instruction, at least for the time being?

Yes.

Thank you!

tkchia:
Hello @jart, hello @ghaerr,

Another minor problem I am thinking about, is whether it will be good to try to protect the "BIOS" code from being overwritten by guest code, and if so, how to do this.

(Ideally the BIOS ROM area will seem to be writable, but writes will have no effect.  The next best thing to do, I suppose, might be to `mprotect(`...`, PROT_READ)` the area &mdash; and unprotect it before freeing `s->real`.  But doing this reliably under differing OS page sizes might be a problem.)

Thank you!

ghaerr:
Hello @tkchia,

> whether it will be good to try to protect the "BIOS" code from being overwritten by guest code, and if so, how to do this.

For `blink16`, I decided to add a capability beyond what the actual hardware can do, similar in concept to ASAN checking. That is, a separate set of R/W flags for each byte of physical memory (`Byte shadowRam[RAMSIZE]`). Then, the `setShadowFlags` function is used by the binary loader (ELKS a.out or DOS) to allow or prohibit read or write access to the text, data/bss and stack sections of the binary. For OS images, the `setShadowCheck` function enables R/W access for all memory. In the ELKS a.out loader:
```
setShadowFlags(0, ES, tseg, fRead); /* text read-only */
```
This mechanism does use twice the real memory, but I thought worth it since it's only 1M for real mode. `blink` has its own implementation of shadow RAM using specially addressed virtual memory which might be able to be used for this case.

> But for something like the int3 handler, we do actually want the hypervisor to see the %cs:%rip at the site of the actual breakpoint, not at the site of the ISR. 

I see. When single stepping through code that executes int3 though, one will still see the `int 3` itself and the following `hvjmp` before magically changing the disasm pane display to the CS:IP of the int3 instruction once the VM is halted. I'm not sure all the trickery is worth it just for int3 display, couldn't the internal `OnHalt` handler adjust to the displayed CS:IP for this single case? It's probably not a big deal leaving it the way you have it working now, but IMO the less magic the better for a superior visually debugging display environment, since both `hvjmp` and `hvcall` are being invented just to allow an external BIOS to be executed.

The possibility of using special memory addresses (e.g. fixed BIOS ROM addresses corresponding to interrupt handlers entry points) comes to mind (perhaps somewhat like how you mentioned dosemu works?) - the IVT could be loaded with FFFF:FExx addresses that point to BIOS ROM and are never actually jumped to but executed similarly to `hvjmp`. This is pretty magic as well, but has the advantage of the emulator not actually executing code outside of the initial INT instruction itself, and no need to setup 256 IVT vectors and 256 `hvjmp` instructions.

> The next best thing to do, I suppose, might be to mprotect(..., PROT_READ) the area 

It occurs to me that one might be able to use another shadow RAM bit or the existing breakpoint mechanism to catch interrupts or not with some automatic monitoring of the IVT area contents being changed - but that's also a lot of work for just getting int3 to display properly and/or to eliminate `hvjmp`.

Thank you!

blink/defbios.c
```diff
@@ -0,0 +1,180 @@
...
+  Put32(s->real + 0x72 * 4, kBiosSeg << 16 | (kBiosDefInt0x72 - kBiosBase));
+  Put32(s->real + 0x73 * 4, kBiosSeg << 16 | (kBiosDefInt0x73 - kBiosBase));
+  Put32(s->real + 0x74 * 4, kBiosSeg << 16 | (kBiosDefInt0x74 - kBiosBase));
+  Put32(s->real + 0x75 * 4, kBiosSeg << 16 | (kBiosDefInt0x75 - kBiosBase));
+  Put32(s->real + 0x76 * 4, kBiosSeg << 16 | (kBiosDefInt0x76 - kBiosBase));
+  Put32(s->real + 0x77 * 4, kBiosSeg << 16 | (kBiosDefInt0x77 - kBiosBase));
+}
+
+void SetDefaultBiosDataArea(struct Machine *m) {
+  SetWriteAddr(m, 0x400, 0x100);
```
ghaerr, line 169: The `SetWriteAddr` function is only used to highlight a single range of memory addresses changed by the last executed instruction for the TUI, and in addition only works properly when called a single time in between instructions. Thus, it seems this call and the call on Line 127 won't work as possibly intended. I'm not sure it should even be called in setup routines, since these routines are not indicative of instruction execution memory change effects.

tkchia:
Hello @ghaerr,

> When single stepping through code that executes int3 though, one will still see the `int 3` itself and the following `hvjmp` before magically changing the disasm pane display to the CS:IP of the int3 instruction once the VM is halted.

I had originally intended to submit, as a separate PR, an optimization for the `int` instruction to immediately trap to the hypervisor if the ISR is just  `hvjmp`.  It turns out that &mdash; as you pointed out &mdash; doing so will actually also improve the user interface.  So I am now adding this optimization as part of this PR.  :ok_hand: 

Thank you!

jart (approved):
Looks good to me. I'll wait for @ghaerr to approve.

ghaerr:
LGTM as well, except I still don't quite understand the intended use of `SetWriteAddr`, as described by my previous comment above.

blink/defbios.c
jart: Yeah I don't think this does anything. Might as well remove it before we merge.

jart:
@tkchia could you please remove the `_Noreturn` annotation from `OpHlt` in machine.h? It causes an error warning for me.

Here's a screenshot of SectorLISP starting up in Blink. As we can see here, symbol table support could be better.

image

blink/defbios.c
jart: Actually, I take it back. This highlight does appear to happen, after stepping two instructions, which is strange.

image

tkchia: Hello @jart, hello @ghaerr,

My main intention was to indicate that the `hvcall 0x19` was having an effect on the contents the emulated memory.

I suppose I should try invoking the `m->system->redraw` hook (either from `defbios.c` or from `blinkenlights.c`) to get Blink to properly redraw everything?

Thank you!

tkchia:
Hello @jart,

> @tkchia could you please remove the _Noreturn annotation from `OpHlt` in machine.h? It causes an error warning for me.

Whoops.  I will make the change.

Hello @ghaerr,

Would you still like me to rename the `hvjmp` instruction to something else?  (I am thinking of using `hvtailcall`.)

Thank you1

blink/defbios.c
ghaerr: > This highlight does appear to happen, after stepping two instructions, which is strange.

The highlight occurs because the `hvjmp` instruction executes `SetDefaultBiosIntVectors`, as well as `SetBiosDataArea`. However, these procedures change data in non-adjacent memory ranges.

> I suppose I should try invoking the m->system->redraw hook (either from defbios.c or from blinkenlights.c) to get Blink to properly redraw everything?

No, the problem isn't a redraw hook. The problem is that `SetWriteAddr` is called from both the above routines, and the resultant requested memory highlighting isn't cumulative, since `SetWriteAddr` only works for the last-set address range specified. Thus, the results are not likely as intended. `blink` can't at this time display multiple non-adjacent memory ranges, as it was designed more for instruction-affected memory display, rather than say, BIOS areas touched by the execution of an INT instruction.

ghaerr:
Hello @tkchia, hello @jart,

> Would you still like me to rename the hvjmp instruction to something else? (I am thinking of using hvtailcall.)

You're the one doing all the work, and I get stuck up on names sometimes lol. Now that I understand it, `hvjmp` is probably fine, but I like `hvtailcall` although it is kind of long (maybe `hvtcall`)?  Perhaps it can be changed if really desired in another PR when the instruction name itself is actually used in an ASM BIOS.

> Here's a screenshot of SectorLISP starting up in Blink. As we can see here, symbol table support could be better.

I have run into this problem in `blink16`, where I realized its going to be a bit complicated to get things completely correct. This is because we're likely going to have to track CS:IP rather than LVA for symbols, and the current routines all just take a flat address. Depending on exactly what program is currently residing at a given memory address, the same LVA can be a different symbol. (I'm not sure this is the exact problem with the display shown, but this will definitely show up).

On another note, IMO we might be better off *not* starting execution in bare metal mode at a ROM address with INT 19h, but instead just starting with the loaded program shown at 7C00:0. This will make more sense to most people, I believe, as it allows them to see the program they're trying to execute, rather than step through an emulated hardware startup. The INT 19h code will still be useful for bare metal programs that want to reboot. What do you think?

Thank you!

blink/defbios.c
ghaerr: > Yes, for some reason the highlighting still only shows up only after I trace over the first boot sector instruction

This may be the result of some complexity in the two slightly different redraw routines, `Redraw(true)` vs `ReactiveDraw()`. I'll have to look more into this, but remember something similar while straightening this all out for `blink16`. The ReactiveDraw routine only redraws when `tuimode` is set and IIRC eventually forced some redraws within the C INT service routines by calling `Redraw(true)` directly. I can look into this further should you need help.

ghaerr:
I'm good to go with the most recent changes, thank you @tkchia!

tkchia:
@ghaerr : thanks!

tkchia:
Hello @ghaerr,

> On another note, IMO we might be better off not starting execution in bare metal mode at a ROM address with INT 19h, but instead just starting with the loaded program shown at 7C00:0. This will make more sense to most people, I believe, as it allows them to see the program they're trying to execute, rather than step through an emulated hardware startup. The INT 19h code will still be useful for bare metal programs that want to reboot. What do you think?

I guess then the question is what Blinkenlights should do if it is to allow alternative BIOS images in the future.  If debugging is to start at `0:0x7c00`, then will that mean that Blinkenlights should try to "implicitly" execute the BIOS's initialization (perhaps until execution reaches `0:0x7c00`)?  Or should it maybe do something else &mdash; in which case, what will be a correct thing to do?

Thank you!

ghaerr:
Hello @tkchia,

Good discussion, I see your points. I'm sure that execution and initialization could become more complicated in the future, when a seperate BIOS image is actually presented to `blink` along with a OS/program image. In that case, it may be well to execute the passed BIOS startup code. I was thinking more of the present, where "the BIOS" isn't separate from `blink` and the user likely expects to see the program entry point displayed at startup. For that matter, perhaps there needs to be an option as to whether a "full trace" of execution is wanted or not. That is, what is actually being debugged, the BIOS or the passed program? If `blink` defaults to an internal (or external) BIOS, perhaps start at 7C00:0 unless a BIOS image is explicitly specified. This can all probably wait until the separate BIOS is actually implemented, and more users use `blink`.

Thank you!

PR #106, opened 2023-04-17
Implement some i186+ instructions, and fix a few others

tkchia:
* Implement `clts` (clear task-switched flag) instruction
* Partly implement `enter` (make stack frame) instruction, for nesting level = 0
* Fix operand size handling for `leave`, `pusha`, and `popa` instructions
* Disassemble (but not yet execute) `sldt`, `str`, `lldt`, `ltr`, `verr`, and `verw` instructions (https://github.com/tkchia/blink/commit/00d8ba089453943ef8e3e25b8c38fe84c43de2ae)

### DIFF
--- a/blink/legacy.c
+++ b/blink/legacy.c
@@ -42,6 +42,21 @@ relegated void OpDecZv(P) {
   }
 }
 
+static relegated void PushaCommon(struct Machine *m, void *b, size_t n) {
+  u32 v;
+  switch (m->mode.omode) {
+    case XED_MODE_REAL:
+      Put16(m->sp, (v = (Get16(m->sp) - n) & 0xffff));
+      break;
+    case XED_MODE_LEGACY:
+      Put64(m->sp, (v = (Get32(m->sp) - n) & 0xffffffff));
+      break;
+    default:
+      __builtin_unreachable();
+  }
+  CopyToUser(m, m->ss.base + v, b, n);
+}
+
 static relegated void Pushaw(P) {
   u16 v;
   u8 b[8][2];
@@ -53,8 +68,7 @@ static relegated void Pushaw(P) {
   memcpy(b[5], m->dx, 2);
   memcpy(b[6], m->cx, 2);
   memcpy(b[7], m->ax, 2);
-  Put16(m->sp, (v = (Read16(m->sp) - sizeof(b)) & 0xffff));
-  CopyToUser(m, m->ss.base + v, b, sizeof(b));
+  PushaCommon(m, b, sizeof(b));
 }
 
 static relegated void Pushad(P) {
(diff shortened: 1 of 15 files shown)


tkchia:
I hope to add some test programs for the instructions that are implemented and/or fixed.  Thank you!

ghaerr:
Hello @tkchia,

This looks great, you're really diving deep into the VM and filling holes :) The test suite code looks fantastic also.

Out of curiosity, is the deleted `memcpy(m->sp, b[3], 2);` in `Popaw` just an optimization because `leave` never uses the saved SP from the stack frame (i.e. no stack switch occurs and the value is always ignored, even though SP was saved by `enter`)?

Thank you!

tkchia:
Hello @ghaerr,

> Out of curiosity, is the deleted `memcpy(m->sp, b[3], 2);` in `Popaw` just an optimization because `leave` never uses the saved SP from the stack frame (i.e. no stack switch occurs and the value is always ignored, even though SP was saved by `enter`)?

(You mean `popaw' and `pushaw'?)  Not only that &mdash; ignoring the pushed value of `sp` is actually more in line with what an actual x86 CPU actually does (as documented in the Intel manuals).

Thank you!

tkchia:
Hmm.  I hope I have nailed down the correct way to determine the address and operand sizes for `pusha` and `popa`.  Apparently a `0x66` prefix will change the size of the things being pushed, while a `0x67` prefix will simply be ignored.

I guess I should also add some checks to check that `pusha` and `popa` work as intended in 32-bit protected mode.

Thank you!

ghaerr:
> Apparently a 0x66 prefix will change the size of the things being pushed, while a 0x67 prefix will simply be ignored.

Wow, interesting. Are you finding this out via the Intel manual or real hardware? Does the push size default to the state of the D-bit (16 vs 32) in the stack or code segment (protected mode only, not real or long mode) and then possibly overwritten?

tkchia:
Hello @ghaerr,

> Are you finding this out via the Intel manual or real hardware?

Both actually.  Occasionally the Intel documentation might be wrong or unclear, so I have to run some code under either Linux/x64 or MS-DOS to see what it actually does.

Thank you!

test/asm/asm.mk
```diff
@@ -51,8 +51,8 @@ o/$(MODE)/test/asm/%.com:						\
 	@echo "o/$(MODE)/blink/blink -m $< || exit" >>$@
 	@echo "echo [test] o/$(MODE)/blink/blink -jm $< >&2" >>$@
 	@echo "o/$(MODE)/blink/blink -jm $< || exit" >>$@
-	@echo "echo [test] o/third_party/qemu/qemu-x86_64 -cpu core2duo $< >&2" >>$@
-	@echo "$(VM) o/third_party/qemu/qemu-x86_64 -cpu core2duo $< || exit" >>$@
+	@echo "echo [test] o/third_party/qemu/4/qemu-x86_64 -cpu core2duo $< >&2" >>$@
```
jart, line 54: Thanks for catching this.

blink/legacy.c
```diff
@@ -68,53 +82,64 @@ static relegated void Pushad(P) {
...
-  memcpy(m->sp, b[3], 4);
   memcpy(m->bx, b[4], 4);
   memcpy(m->dx, b[5], 4);
   memcpy(m->cx, b[6], 4);
   memcpy(m->ax, b[7], 4);
 }
 
 relegated void OpPusha(P) {
-  switch (Eamode(rde)) {
+  switch (m->mode.omode) {
```
jart, line 135: Why aren't you using `Eamode()`? Don't you want to support the operand-size override prefix?

jart (commented):
Late to the review. See comments.

test/asm/asm.mk
tkchia: @jart : thanks!

blink/legacy.c
tkchia: Hello @jart,

I found that `Eamode()` is not realy the correct property to use here &mdash; specifically, it is controlled by the "address size" prefix (`0x67`) rather than the "operand size" prefix (`0x66`).

Thank you!

tkchia: I believe using `kStackOsz[]` to determine the operand size will (also) do the correct thing.  But `kStackOsz[]` happens to be in `stack.c` and not `legacy.c`.

Thank you!

PR #108, opened 2023-04-19
Fix build on Android

trungnt2910:
- Added `./configure`-time checks for certain POSIX functions that are missing from `bionic` libc.
- Fixed build on `clang`.

Related: termux/proot-distro#281.


jart (approved):
Outstanding work!

PR #109, opened 2023-04-20
Tweak handling of guest machine execution mode

tkchia:
- Split machine mode `m->mode` into two parts, to explicitly distinguish real mode & 16-bit protected mode
  - `m->mode.omode` now says how instructions should be decoded
  - `m->mode.genmode` says whether guest is in real mode or not
- Use `Mode(rde)` more often, so we can optimize away dead code on `--disable-metal`
- Setting `CR0_PE` in guest `%cr0` now immediately triggers machine mode change

### DIFF
--- a/blink/x86.h
+++ b/blink/x86.h
@@ -29,6 +29,30 @@
 #define XED_ILD_MAP6    6
 #define XED_ILD_BAD_MAP 7
 
+#define XED_GEN_MODE_REAL      0
+#define XED_GEN_MODE_PROTECTED 1
+#define XED_GEN_MODE_V86       2  // unimplemented
+
+struct XedMachineMode {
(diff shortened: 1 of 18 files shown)


blink/x86.h
```diff
@@ -29,6 +29,30 @@
 #define XED_ILD_MAP6    6
 #define XED_ILD_BAD_MAP 7
 
+#define XED_GEN_MODE_REAL      0
+#define XED_GEN_MODE_PROTECTED 1
+#define XED_GEN_MODE_V86       2  // unimplemented
+
+struct XedMachineMode {
+  uint8_t omode : 2,    // bitness of operands, for instruction decoding
```
jart, line 37: Bitfields tend to be a little janky in my experience but let's try it.

jart (approved):

jart:
Thank you!

PR #114, opened 2023-04-24
Various minor changes

tkchia:
* Recognize upper case `.BIN`, `.IMG`, `.RAW`, & `.COM` file extensions
* Improve display of segment registers
* Use `0xF000:0xFFF0`, not `0xFFFF:0`, as metal mode entry point


jart (approved):
Thank you!

PR #121, opened 2023-04-30
Add CGA 40x25 and BIOS Data Area console variables

ghaerr:
Adds support for BIOS CGA modes 0 and 1 (40x25 text). 
All CGA/MDA variables are now stored in the BIOS Data Area for strict IBM PC compatibility.
Adds `Bda*` macros for accessing BIOS Data Area variables in `blink/bda.h`.
Implements cursor hide function.
Implement several more `int 10h` BIOS console functions.
Tracked down and fixed several bugs/crashes in BIOS console emulation, `unassert` macros left in.
R)eset now operates properly and resets the BIOS or pty screen when restarting the OS/boot emulation.
V)mode flip between BIOS mode 3 (CGA 80x25) and pty now works, which allows OS or boot sector startup in either of two modes, which is useful when the default BIOS mode 3 isn't desired.

Its now possible to boot a [TetrOS boot sector game](https://github.com/daniel-e/tetros/tree/master) which runs in CGA 40x25 mode. Unfortunately, it uses the arrow keys for moving the tetris blocks around and `int 16h` doesn't return scan codes yet so the user always loses.

Plans for the future are to setup BDA variables that might allow the ELKS direct console to run, which doesn't make BIOS calls but instead draws directly to video ram and talks to the 6845 CRT controller for cursor management.

Fairly heavily tested on ELKS, FreeDOS, SectorLISP, and TetrOS.

Here's TetrOS running in CGA 40x25 mode:
TetrOS blink

### DIFF
--- a/blink/bda.h
+++ b/blink/bda.h
@@ -1,0 +1,39 @@
+#ifndef BLINK_BDA_H_
+#define BLINK_BDA_H_
+#include "blink/machine.h"
+#include "blink/endian.h"
+
+#define BDA(address)        (g_machine->system->real + address)
+
+#define    BdaVmode         Read8(BDA(0x0449))
+#define SetBdaVmode(v)      Write8(BDA(0x0449),v)
+#define    BdaCols          Read8(BDA(0x044A))
+#define SetBdaCols(v)       Write8(BDA(0x044A),v)
+#define    BdaPagesz        Read8(BDA(0x044C))
+#define SetBdaPagesz(v)     Write8(BDA(0x044C),v)
+#define    BdaCurx          Read8(BDA(0x0450))
+#define SetBdaCurx(v)       Write8(BDA(0x0450),v)
+#define    BdaCury          Read8(BDA(0x0451))
+#define SetBdaCury(v)       Write8(BDA(0x0451),v)
+#define    BdaCrtc          Read16(BDA(0x0464))
+#define SetBdaCrtc(v)       Write16(BDA(0x0464),v)
+#define    BdaCurend        Read8(BDA(0x0460))
+#define SetBdaCurend(v)     Write8(BDA(0x0460),v)
+#define    BdaCurstart      Read8(BDA(0x0461))
+#define SetBdaCurstart(v)   Write8(BDA(0x0461),v)
+#define BdaCurhidden        ((BdaCurstart & 0x20) || (BdaCurstart > BdaCurend))
+
+#if 0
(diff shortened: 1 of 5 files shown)


ghaerr:
Does anyone have thoughts on why `note_test.com` is failing with VFS enabled on the automated run? I'm not familiar with how to run the test suites locally to try to repeat this. The changes in this PR should not affect `o//blink/blink` runs.

tkchia:
Hello @ghaerr,

> I'm not familiar with how to run the test suites locally to try to repeat this.

You can run
```
make check
```
for basic tests &mdash; this will also try to download some code that is needed to run the tests.  Run
```
make check2
make emulates
```
for additional tests.

I generally try to add additional tests to `test/metal/` when I add new stuff to Blink on the bare metal side.  In my view it is worth putting in a bit of extra time and effort to maintain code quality (since this project is not exactly in a race against time (or against anything else)).

Thank you!

tkchia:
The 40 &times; 25 text video mode support seems to me like a missed opportunity to play with Unicode double width characters (partially implemented in `blink/pty.c`).  Admittedly though, one challenge will be to get the full set of CP437 characters to display in some double-width form, including things like ☻ and δ.

Thank you!

ghaerr:
Hello @tkchia,

Thanks for the information on the local tests. Unfortunately, `make check` dies on a number of tests on macOS, including for instance `popcnt_test.com`, which have nothing to do with this PR. The good news is, for some reason, this PR is now showing all tests passed (I'm not sure how that happened, as I didn't see a button to rerun the tests). It seems there might be some flakiness in the GH autorun test runner?

> The 40 × 25 text video mode support seems to me like a missed opportunity to play with Unicode double width characters 

Now that you mention it, it remember noticing that QEMU is drawing double width characters in this mode. Of course, adding the capability to double-draw each character dot in a GUI is lots easier than our TUI. I have considered the idea of popping up an SDL graphics console within `blink` for that sort of thing, I have the code ready to go, but don't like a required dependency on SDL. I only added 40x25 mode because I wanted to test `TetrOS` as a boot sector game and it required mode 0. Frankly, I doubt anyone will ever really use 40x25 in earnest, but I suppose it would be fairly straightforward to add a new routine `DrawCga40` and an additional translation table for CP 437 -> Unicode to handle the case.


Given that this PR contains a few other enhancements and some bug fixes, I would like to merge it now, with true double-wide 40x25 considered in another submission, if that's ok with you?

Thank you!

tkchia (approved):

PR #123, opened 2023-05-01
Tweak `int3`/`int` opcode handling, & fix display-related issues

tkchia:
- Do not halt guest execution on `int3` if guest has installed an `int3` handler
- In general, treat each `int`, `int3`, or `into` instruction as a sort of call that may return to its following instruction
- Fix some display-related issues


jart (approved):

jart:
Thank you!

PR #124, opened 2023-05-01
More fixes to BIOS INT 0x10 routines

ghaerr:
Implements BDA and BIOS fixes to allow FreeDOS install to display more accurately.

`VFRAME.COM` now works.

Rewrote `int 0x10` function 0x09 and added function 0x0A.

Implements EGA BIOS `Lines` variable at BDA+0x484 in order for FreeDOS and `vframes` to work. Frankly, I'm amazed at the amount of peeking and poking into the BDA that's done by DOS programs in order to operate correctly.

Removes various `unassert`s - turns out many DOS programs position the cursor off the screen and write. Real PC BIOS (and QEMU) don't do bounds checking and just index and draw into video RAM even though it may not be displayed. Ever.

Other `unassert`s left in, but these may also want to be removed in the future. For now, leaving them in gives some semblance of bounds checking, until another DOS program does something funky. At least we'll know about it.

Here is the now-correct FreeDOS installation screen after discontinuing install with `N`:

FreeDOS blink OK

~~The only remaining problem with the FreeDOS installer is the reverse-video highlighting displayed during language selection. It must be doing something quite unnatural, as I haven't been able to figure that one out yet.~~ This was fixed by not writing the attribute byte in `int 0x10` function 0x0E.

### DIFF
--- a/blink/bda.h
+++ b/blink/bda.h
@@ -7,10 +7,12 @@
 
 #define    BdaVmode         Read8(BDA(0x0449))
 #define SetBdaVmode(v)      Write8(BDA(0x0449),v)
-#define    BdaCols          Read8(BDA(0x044A))
-#define SetBdaCols(v)       Write8(BDA(0x044A),v)
-#define    BdaPagesz        Read8(BDA(0x044C))
-#define SetBdaPagesz(v)     Write8(BDA(0x044C),v)
+#define    BdaCols          Read16(BDA(0x044A))         // returns actual # columns
+#define SetBdaCols(v)       Write16(BDA(0x044A),v)      // BIOS stores # columns
+#define    BdaLines         (Read8(BDA(0x0484)) + 1)    // returns actual # lines
+#define SetBdaLines(v)      Write8(BDA(0x0484),v)       // BIOS stores # lines - 1
+#define    BdaPagesz        Read16(BDA(0x044C))
+#define SetBdaPagesz(v)     Write16(BDA(0x044C),v)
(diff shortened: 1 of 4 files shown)


ghaerr:
Hello @tkchia,

I'm continuing to get build failures in the auto-test runner again, can you see them? It's failing in `make matrix` on `nanosleep_test.com`. Not sure what to do. I am also unable to restart the test.

Thank you!

ghaerr:
Hello @tkchia,

After having finally figured out the last FreeDOS display problem (i.e. write teletype should not write any attribute passed in BL), I pushed the fix and the tests passed this time. There is obviously something flakey with the CI test suites, but I don't have any idea what that might be.

There are now no known display problems with ELKS, FreeDOS or SectorLISP.  Here's the first FreeDOS boot screen finally displaying properly, using it's intermixed `int 10h` calls and accessing video ram directly:
FreeDOS blink OK 1

Thank you!

tkchia (approved):

tkchia:
Hello @ghaerr,

> What do you think of this split-out idea, and would you mind if I took a pass at doing it?

Me, personally I should be OK with this.  In fact I had been thinking myself of factoring out the various BIOS software interrupts (video, disk, etc.) into their own separate modules, though I never got around to doing so.

Thank you!

tkchia:
Hello @ghaerr,

Personally I would also suggest to do the refactoring in several small steps, rather than to try to do one mega-restructuring (which might be confusing and also risky).

Thank you!

PR #125, opened 2023-05-02
Separate C BIOS handlers into bios.c from blinkenlights.c

ghaerr:
Moves BIOS interrupt handling routines out of `blinkenlights.c` into new `bios.c` file, requested in https://github.com/jart/blink/pull/124#issuecomment-1531661640.

As recommended by @tkchia, this is a first pass and only repositions source code, no other changes. In the future, some `extern` declarations may want to be changed or better integrated.

All-in-all, its pretty nice to see how independent these files are, with the exception of just a small handful of shared variables and callback functions. As previously mentioned, this separation will allow for better up and down stream integration of enhancements to/from `blink` and `blink16`, as well as keep the TUI itself more isolated from the specific architecture(s) it operates. We may also want to consider moving the console display functions (`DrawDisplay`, `DrawDisplayOnly` and `DrawTerminal` out of blinkenlights.c.

### DIFF
--- a/blink/bios.c
+++ b/blink/bios.c
@@ -1,0 +1,941 @@
+/*-*- mode:c;indent-tabs-mode:nil;c-basic-offset:2;tab-width:8;coding:utf-8 -*-│
+│vi: set net ft=c ts=2 sts=2 sw=2 fenc=utf-8                                :vi│
+╞══════════════════════════════════════════════════════════════════════════════╡
+│ Copyright 2023 Justine Alexandra Roberts Tunney                              │
+│                                                                              │
+│ Permission to use, copy, modify, and/or distribute this software for         │
+│ any purpose with or without fee is hereby granted, provided that the         │
+│ above copyright notice and this permission notice appear in all copies.      │
+│                                                                              │
+│ THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL                │
+│ WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED                │
+│ WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE             │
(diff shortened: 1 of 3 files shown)


ghaerr:
It seems the checks failing this time around are possibly due to missing include files used by different compilation configurations. Unfortunately, I am only able to compile and run `blink` with very minimal options (`./configure --disable-all --enable-metal`) on macOS 10.15.7, using `clang` v11.0.0, as previously described. Thus, I am regretfully unable to complete the error-checking involved for the various Blink configuration modes.

Please let me know if I should delete this and give up, or whether there might be ways I can execute `make matrix` or otherwise emulate the GitHub CI on macOS.

tkchia:
Hello @ghaerr,

The particular [error](https://github.com/jart/blink/actions/runs/4866810935/jobs/8678720288?pr=125) from GitHub CI seems to be because `blink/overlays.h` was not included.  Perhaps the correct thing to do is to tweak `blink/vfs.h` to also include `blink/overlays.h`...

> When compiling using `./configure` with no options on macOS, the `clang` compiler actually segfaults on `blink/signal.c`; I have not been able to locate the offending line(s) yet.

This is not cool.  Maybe also try upgrading the `clang` installation on your Mac?

Thank you!

ghaerr:
Hello @tkchia,

Thank you for your suggestions and help, they are much appreciated. I am glad to see your suggestion worked, as I tried reconfiguring with `--enable-vfs` and there are more problems with compilation with that option enabled on macOS, including `atomic_exchange` and `fdatasync` missing, and `signal.c` still segfaulting the compiler. Those issues can be dealt with at another time, I suppose.

At the moment, I don't see an immediate need for more segregation of `bios.c` into other files, but this should help us continue to add BIOS features away from the TUI.

Thank you!

tkchia:
Hello @ghaerr,

I am working on a patch to add `fdatasync()` to the `./configure` checks.  I am not sure what to do about the `atomic_exchange()` calls, but we will probably figure it out.

Thank you!

PR #130, opened 2023-05-07
[metal] Tweak display of CP437 characters in CGA/MDA window

tkchia:
* display ASCII characters 0x20--0x7E as-is
* display 0xFF as less intrusive symbol for No Break Space
* display 0x7F as "house", following actual hardware, rather than passing verbatim to host TTY

### DIFF
--- a/blink/bios.c
+++ b/blink/bios.c
@@ -558,21 +558,6 @@ static void OnVidyaServiceGetCursorPosition(void) {
   }
 }
 
-static int GetVidyaByte(unsigned char b) {
-  if (0x20 <= b && b <= 0x7F) return b;
-#if 0
-  /*
-   * The original hardware displayed 0x00, 0x20, and 0xff as space. It
-   * made sense for viewing sparse binary data that 0x00 be blank. But
-   * it doesn't make sense for dense data too, and we don't need three
-   * space characters. So we diverge in our implementation and display
-   * 0xff as lambda.
-   */
-  if (b == 0xFF) b = 0x00;
-#endif
-  return kCp437[b];
-}
-
 static void OnVidyaServiceReadCharacter(void) {
   u16 *vram;
   u16 chattr;
--- a/blink/blinkenlights.h
+++ b/blink/blinkenlights.h
@@ -4,7 +4,11 @@
 
 #include <stdbool.h>
 #include <stdlib.h>
+#include <wchar.h>
 #include <sys/types.h>
+
+#include "blink/likely.h"
+#include "blink/util.h"
 
 #define kModePty 255
 
@@ -27,5 +31,16 @@ ssize_t ReadAnsi(int fd, char *p, size_t n);
 /* bios.c */
 void VidyaServiceSetMode(int);
 bool OnCallBios(int interrupt);
+
+static inline wint_t GetVidyaByte(unsigned char b) {
+  if (LIKELY(0x20 <= b && b <= 0x7E)) return b;
+  /*
+   * In the emulated screen, show 0xff as a "shouldered open box"
+   * instead of lambda.  The Unicode 4.0 charts tell us that this
+   * glyph is an ISO 9995-7 "keyboard symbol for No Break Space".
+   */
+  if (UNLIKELY(b == 0xFF)) return 0x237D;
+  return kCp437[b];
+}
 
 #endif /* BLINK_BLINKENLIGHTS_H_ */
--- a/blink/cga.c
+++ b/blink/cga.c
@@ -22,6 +22,7 @@
 #include <stdio.h>
 #include <termios.h>
 
+#include "blink/blinkenlights.h"
 #include "blink/buffer.h"
 #include "blink/macros.h"
 #include "blink/util.h"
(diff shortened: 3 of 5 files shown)


tkchia:
image

ghaerr:
Hello @tkchia,

> display ASCII characters 0x20--0x7E as-is

Agreed, already does.

> display 0xFF as less intrusive symbol for No Break Space

If the object is to accurately display the CGA/MDA screens as the IBM PC does, why are we changing the display of 0xFF to something other than space (blank)?  There could be any number of PC programs that use character 0xFF to display blank, does the real GWBASIC on a PC actually display a glyph that is not blank? (I think not. I haven't yet tried running GWBASIC to see it's startup screen, since the only copy I have is tied to FreeDOS v1.3-rc1, which now won't run on `blink` without commenting out the new `Mprotect` code).

> display 0x7F as "house", following actual hardware, rather than passing verbatim to host TTY

Agreed. The old code already does.

Unless we really want to diverge from displaying a compatible blank space for 0xFF, the only actual change required for this PR is the attribute fix on line 65 of blink/cga.c (and correspondingly for blink/mda.c). Adding an extra procedure call and three more if checks may also slow down our display considering it may be updated after every instruction in some modes... 

Thus, I'm not really in agreement with this PR. (The if check on line 562 of blink/blinkenlights.c could stay for the benefit of the PTY output, though, and still keep the CGA/MDA redisplay very fast without calling `GetVidyaByte` needlessly).

Thank you!

tkchia:
Hello @ghaerr,

> If the object is to accurately display the CGA/MDA screens as the IBM PC does, why are we changing the display of 0xFF to something other than space (blank)?

Well &mdash; this is @jart 's project.  Presumably she decided to write
```
#if 0
  ...
  if (b == 0xFF) b = 0x00;
#endif
```
for a reason.  Maybe we should honour that.

Thank you!

tkchia:
Hello @ghaerr,

> Adding an extra procedure call and three more if checks may also slow down our display considering it may be updated after every instruction in some modes...

What is the exact performance impact?  Do we even know it?  Anyway, if a procedure call is too slow I suppose we can try to turn it into an inline function.  Let us figure out the correct thing to do, then figure out how to do it fast.

Thank you!

ghaerr:
Hello @tkchia,

> Let us figure out the correct thing to do, then figure out how to do it fast.

Agreed. I suppose my last post was a result of being a bit frustrated, no worries :)

> Presumably she decided to write ... for a reason. Maybe we should honour that.

We are discussing two different things here: the referenced code was for the PTY driver, as the original CGA/MDA code never called it. It may make some sense for PTY output to differentiate between various character values instead of mapping three versions to `<space>`. However, the PTY driver cannot be used for CGA/MDA output, and the CGA/MDA driver, when I started working on it, did not work well. So I enhanced it to actually produce what is seen on an IBM CGA display. I believe we should be accurate in our CGA emulation, and not stick to code that was written for PTY display, as they're just not compatible anyways. My frustration is that there apparently isn't agreement as to whether we are trying for accuracy in display emulation, or something else. The PTY driver is a greenfield invention and doesn't have to be compatible (and isn't). That's one of the reasons I didn't consider using it when getting the CGA display working properly.

Thank you!

tkchia:
Hello @ghaerr,

> the referenced code was for the PTY driver, as the original CGA/MDA code never called it

`GetVidyaByte()` was specifically used for `int 0x10` calls (and not e.g. terminal emulation for ELF programs).  Thank you!

ghaerr:
@tkchia: 

> GetVidyaByte() was specifically used for int 0x10 calls (and not e.g. terminal emulation for ELF programs).

Yes, but never originally for any BIOS modes 2, 3 or 7, as `DrawCga` or `DrawMda` were called instead. Originally, the PTY driver would be used if a BIOS set mode call wasn't explicitly made, but that doesn't work properly for IBM PC display modes. In the new code, real mode initialization always sets mode 3, and the PTY driver will never be used. There could be an issue with this if an ELF program can run in real mode.

Can an ELF program run in real mode? If so and it executes `int 0x10`, then `m->metal` won't be set and the PTY driver will be used which calls `VidyaGetByte` and in that case 0xFF -> lambda. Otherwise, the new BIOS console output rewrite will never use PTY unless 'v' is pressed (which is mostly useless but still possibly useful for seeing output run through the PTY driver, I suppose).

Thank you!

tkchia:
Hello @ghaerr,

> > display ASCII characters 0x20--0x7E as-is

> Agreed, already does.

Not quite true actually.  In particular the original `cga.c` and `mda.c` code was rendering `0x3f` as `⁇` rather than `?`.  This is noticeable when you run `sectorlisp-friendly.bin`, for example.  `0x22` and `0x27` were also specially rendered as `“` and `‘`.  I believe the intention was for these to be rendered as normally ASCII characters.

Thank you!

ghaerr:
Hello @tkchia,

> the original cga.c and mda.c code was rendering 0x3f as ⁇ rather than ?.
> for example.  0x22 and 0x27 were also specially rendered as “ and ‘

I'd seen the ⁇ display and knew it was coming from the `kCp437[]` translation, but now am realizing the likely point of that conversion table isn't for accurate CGA display, but rather for the better memory contents display scanning. As such, this obviates the need for both a `GetVidyaByte` function call as well as an additional compare for `0xff`, for accurate CGA display.

As far as inaccurately displaying `0xff` as a non-blank character, I don't really understand why we'd want that, especially with each of GWBASIC's prompts showing it. But if that's what you want, lets go with it, as its easily changed. (Why GWBASIC even writes `0xff` to the console to display a blank space on the PC in the first place is something I have not figured out).

Thank you!

ghaerr (approved):

PR #131, opened 2023-05-07
[metal] Ignore writes to BIOS code area

tkchia:
Also add a `--disable-rom` build option which will effectively make the BIOS code area read/write, like normal RAM.  This shaves around 200 bytes off a `MODE = tiny` Blink executable, at the cost of breaking emulation fidelity.

### DIFF
--- a/blink/memory.c
+++ b/blink/memory.c
@@ -23,6 +23,7 @@
 #include <sys/mman.h>
 
 #include "blink/assert.h"
+#include "blink/biosrom.h"
 #include "blink/bus.h"
 #include "blink/checked.h"
 #include "blink/debug.h"
@@ -445,13 +446,16 @@ u8 *ReserveAddress(struct Machine *m, i64 v, size_t n, bool writable) {
   }
 }
 
-u8 *AccessRam(struct Machine *m, i64 v, size_t n, void *p[2], u8 *tmp,
-              bool copy) {
+static u8 *AccessRam2(struct Machine *m, i64 v, size_t n, void *p[2], u8 *tmp,
+                      bool copy, bool protect_rom) {
   u8 *a, *b;
   unsigned k;
   unassert(n <= 4096);
   if ((v & 4095) + n <= 4096) {
-    return ResolveAddress(m, v);
+    a = ResolveAddress(m, v);
+    if (!protect_rom || !IsRomAddress(m, a)) return a;
+    if (copy) memcpy(tmp, a, n);
+    return tmp;
   }
   STATISTIC(++page_overlaps);
   k = 4096;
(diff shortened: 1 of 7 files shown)


ghaerr:
Hello @tkchia,

I'm still a bit confused here... if `DISABLE_ROM` is not defined, why is the `Mprotect` code still required? Won't that cause a segfault? I thought that this new code effectively checks for and ignores writes within the ROM region if not disabled, thus removing the need for any memory protection kernel services. (And thus, the emulation would in fact be entirely accurate - no segfault or halting the emulator, and writing to ROM locations doesn't alter memory contents).

Thank you!

tkchia:
Hello @ghaerr,

> I thought that this new code effectively checks for and ignores writes within the ROM region if not disabled, thus removing the need for any memory protection kernel services.

My (semblance of a) plan is to use `AccessRam()` to catch most attempts to write to ROM, and rely on the `Mprotect()` to catch any remaining write attempts that may have been missed.  This way, we can find and fix them.

In particular, the `Mprotect()` alerted me to the fact that the "enhanced" `rep stosb` implementation was bypassing the whole `AccessRam()` mechanism.

Thank you!

tkchia:
@ghaerr : I am thinking of turning the `IsRomAddress()` into an inline function, perhaps in `blink/biosrom.h`.  This should hopefully make everything more efficient (and still correct).

ghaerr:
@tkchia,

> I am thinking of turning the IsRomAddress() into an inline function, perhaps in blink/biosrom.h. This should hopefully make everything more efficient (and still correct).

That's a good idea.

> In particular, the Mprotect() alerted me to the fact that the "enhanced" rep stosb implementation was bypassing the whole AccessRam() mechanism.

I see. OK.

This looks fine, and shouldn't probably reduce the execution speed measurably. I'll plan on running `./configure --disable-rom` so I can continue to use FreeDOS v1.3-rc1.

Thank you!

blink/memory.c
```diff
@@ -445,13 +446,31 @@ u8 *ReserveAddress(struct Machine *m, i64 v, size_t n, bool writable) {
...
-              bool copy) {
+#ifndef DISABLE_ROM
+bool IsRomAddress(struct Machine *m, u8 *r) {
+  if (m->metal) {
+    u8 *real = m->system->real;
+#if CAN_ABUSE_POINTERS
+    ptrdiff_t d = r - real;
+    if (kBiosOptBase <= d && d < kBiosEnd) return true;
+#else
+    if (real[kBiosOptBase] <= r && r < &real[kBiosEnd]) return true;
```
ghaerr, line 457: @tkchia: I just noticed this here: I think you mean `&real[kBiosOptBase]` on the left hand side?

tkchia:
@ghaerr : FreeDOS 1.3-rc1 should now work under either `--disable-rom` or `--enable-rom`.  Thank you!

tkchia:
Hello @ghaerr,

> So the idea of leaving in `Mprotect` is to catch internal coding issues at this point.

Yep.  Maybe I should clarify this in the code comments.

Let me also see if I can add a test program, before I merge this PR.

Thank you!

tkchia:
I change my mind &mdash; I will write the test case later.  While trying to come up with a test case I hit upon some other emulator issues, and I hope to sort them out first.

PR #132, opened 2023-05-09
Fix issues related to SIMD opcodes, `rep movsb`, & memory protection

tkchia:


ghaerr:
Hello @tkchia,

It occurred to me while reviewing this code that when `--enable-rom` is configured, `blink` will dutifully ignore memory writes to the BIOS ROM address range, which is nice except that now developers and/or testers won't actually learn that the emulated program possibly tried to do something likely not wanted. Perhaps `IsRomAddress` could make that information known, if desired (when returning `true`): it could `LOGF` the write address being ignored, along with the disassembled instruction and/or IP, or it could halt the machine by throwing an `OnHalt` interrupt with a predefined, say `kRomWrite` error. (Or do nothing, as it does now). I realize that might require a command line option, or even perhaps just a compiled-in flag that could be set by developers that wanted to have the ability to learn more about what the program is doing. Just a thought!

Thank you!

tkchia:
@ghaerr : `LOGF` or `OnHalt` sounds like an idea.  Meantime, I hope to get the correctness issues sorted out first.  The x86 instruction set is large and confusing.  Thank you!

PR #136, opened 2023-05-17
Implement single-stepping (int 1 upon TF = 1)

tkchia:


ghaerr:
Ditto my comment on https://github.com/jart/blink/pull/135#issuecomment-1552148471.

tkchia:
Hello @ghaerr,

Looks like the tests passed when I retried them.  :raised_eyebrow:

ghaerr:
That's great @tkchia. Now if we could just figure out why they're inconsistently failing... and I'm still wondering about `popcnt`, since it failed on me twice without any changes to the VM!

tkchia:
Hello @jart,

I would like your feedback on this PR, as it is somewhat intrusive and impinges on the underlying workings of Blink's JITter.

My general goal is to keep the common case &mdash; x86-64 guest code running with JIT enabled, and without single-stepping &mdash; as fast as before.

Thank you!

jart:
I'll take a look into this in a week when I'll have time to give this the attention it needs.

PR #138, opened 2023-05-20
Special-case `rep ret` to return flags to caller routine

tkchia:
This allows UPX's (https://upx.github.io/) x86-64 unpacker routines to work properly.  UPX relies on a callee being able to return a carry flag to its caller, and also just _happens_ to use `rep ret`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1005,6 +1005,9 @@ carry flag as a return value to indicate error. This should work fine
 when `STC` is used to set the carry flag, but if the code computes it
 cleverly using instructions like `SUB`, then EFLAGS might not change.
 
+As a special case, if a `RET` instruction sports a `REP` prefix, then
+Blink can return flags across the `RET`.
+
 ### Faults
 
 Blink may not report the precise program counter where a fault occurred
--- a/blink/deps.c
+++ b/blink/deps.c
@@ -85,6 +85,11 @@ int GetFlagClobbers(u64 rde) {
       return 0;
     case 0xE8:   // call
     case 0xC3:   // ret
+      if (Rep(rde)) {
+        return 0;
+      } else {
+        return -1;
+      }
     case 0x105:  // syscall
       return -1;
     case 0x000:  // add byte


jart (approved):

ghaerr:
Hello @tkchia,

Agreed this is a pretty brilliant way of handling the problem of UPX's requirement, without adding overhead.

I am curious though: what exactly does `rep ret` do on real hardware, and why does UPX *just happen* to use it?

Thank you!

tkchia:
Hello @ghaerr,

> what exactly does `rep ret` do on real hardware, and why does UPX _just happen_ to use it?

`rep ret` mainly does the same thing as a `ret`.  It was [reportedly](https://repzret.org/p/repzret/) useful for working around a performance issue on some older AMD processors.  GCC will actually emit `rep ret` under certain conditions:
```
$ cat test5.c
#include <inttypes.h>

int y;

uint8_t
f (void)
{
  int x = y;
  return x == 2 ? 1 : x * x;
}
$ gcc -S -O3 test5.c -fno-pic -static -fomit-frame-pointer -mtune=k8
$ cat test5.s
	.file	"test5.c"
	.text
	.p2align 4
	.globl	f
	.type	f, @function
f:
.LFB0:
	.cfi_startproc
	endbr64
	movl	y(%rip), %edx
	movl	$1, %eax
	cmpl	$2, %edx
	je	.L1
	movl	%edx, %eax
	imull	%edx, %eax
.L1:
	rep ret
	.cfi_endproc
...
```
The UPX decompression routines are partly hand-coded in assembly, but presumably the authors also had this AMD performance issue in mind.

Thank you!

ghaerr:
Hello @tkchia,

Thank you for your explanation, quite interesting. It's kind of amazing what the CPU designers are doing for branch prediction optimizations and CPU pipelines after the CISC instruction set is defined. However, I am coming closer to the realization that nothing should surprise me when it comes to the (over)complexity of AMD and Intel x86 processors!!

PR #142, opened 2023-06-05
Update README.md

japrozs:


tkchia:
Hello @jart, hello @ghaerr,

I have not quite been able to verify whether `gmake` works better on macOS than just plain `make` &mdash; or indeed whether it works at all.  What is your experience?

Thank you!

ghaerr:
Hello @tkchia,

> I have not quite been able to verify whether gmake works better on macOS than just plain make

Currently, `Makefile` requires that a version of `make` greater than v3.8.1 be used. When [homewbrew is used to install make](https://stackoverflow.com/questions/43175529/updating-make-version-on-mac), `make` remains at v3.8.1, and `gmake` runs v4.3. Thus, the real issue is not so much that `gmake` is required, instead it's the default from a homebrew install on macOS.

The make bug was previously-reported in [make v3.8.1 bugs building Cosmopolitan](https://github.com/jart/cosmopolitan/issues/375), where `$(file ...)` is silently ignored and `make` version v4.3 fixes that issue.

Thank you!

ryandesign:
It has nothing to do with Homebrew. In macOS Apple ships GNU make 3.81 because it is the last version licensed GPL 2. GNU make 3.82 and later are licensed GPL 3 and Apple will not ship any software licensed GPL 3.

PR #150, opened 2023-08-09
blink/checked.h: fix syntax error when building with GCC 9.4

tkchia:
```
blink/checked.h:10:47: error: missing binary operator before token "("
   10 |      (defined(__has_builtin) && (__has_builtin(__builtin_add_overflow) && \
      |                                               ^
```

### DIFF
--- a/blink/checked.h
+++ b/blink/checked.h
@@ -1,23 +1,41 @@
 #ifndef EASYCKDINT_H_
 #define EASYCKDINT_H_
-#if defined(__has_include) && __has_include(<stdckdint.h>)
+
+#ifdef __has_include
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

>  I think it is generally a good idea for code to test directly for features they want, whenever possible, in contrast to testing version numbers. 🙂

I only ask that because, in the lines immediately before the `__has_builtin` checks, we check for compiler version numbers!! Is that version check for gcc necessary, as we're essentially duplicating the definitions of `chk_add` et al below.

Thanks for the information on the safe integer proposal in C23.

Thank you!

PR #160, opened 2023-12-11
Log: print ellipsis as intended if the log message is truncated

dottedmag:
n is passed to WriteError(), so if it is decremented the ellipsis is not going to be printed.


jart (approved):
Thanks!

PR #161, opened 2023-12-11
APPEND: avoid overrunning the buffer

dottedmag:
if two APPENDs are called one after another, and first of them cannot fit the message into the buffer, the second will end up writing its message pass the end of the buffer (due to size parameter to snprintf being unsigned).


dottedmag:
I'm not entirely sure the checks are needed in all the APPEND variants.

jart (approved):
Thanks!

PR #184, opened 2025-03-11
Add host architecture to cpuid

bjia56:
This change adds reading the host architecture to blink's custom cpuid instruction. With this, an x86_64 application can query the underlying architecture to perform architecture-specific logic, like extract arch native binaries out of a Cosmopolitan binary's zipos.


jart (changes requested):
Mind updating the README?

jart (approved):

PR #185, opened 2025-03-13
Use PROT_MPROTECT for NetBSD mprotect restrictions

bjia56:
NetBSD's mprotect seems to be more restrictive than other OSes, where it does not allow less restrictive mappings than the original mmap, and can be subject to PaX restrictions. According to NetBSD documentation, there is a PROT_MPROTECT macro to define allowed protections for later uses of mprotect, without granting the permissions immediately in mmap. This can be used to ensure that the full range of protections blink could make in the course of execution are permitted.

### DIFF
--- a/blink/map.c
+++ b/blink/map.c
@@ -169,6 +169,11 @@ void *Mmap(void *addr,     //
 #if LOG_MEM
   char szbuf[16];
 #endif
+#if defined(__NetBSD__)
+  if (!(flags & MAP_SHARED)) {
+    prot |= PROT_MPROTECT(PROT_EXEC | PROT_WRITE | PROT_READ);
+  }
+#endif
   res = PortableMmap(addr, length, prot, flags, fd, offset);
 #if LOG_MEM
   FormatSize(szbuf, length, 1024);


jart:
What NetBSD architecture? I don't get this behavior on x86-64.

bjia56:
I get this on both x86_64 and aarch64, NetBSD 10.0 running in VMs

bjia56:
My setup:
```
vmactions# uname -a 
NetBSD vmactions.org 10.0 NetBSD 10.0 (GENERIC) #0: Thu Mar 28 08:33:33 UTC 2024  mkrepro@mkrepro.NetBSD.org:/usr/src/sys/arch/amd64/compile/GENERIC amd64
```

Building the latest code from master with `./configure --static`, then running:
```
vmactions# o/blink/blink third_party/cosmo/tinyhello.elf 
assertion failed
blink/loader.c:71:3146 assertion failed: !ProtectVirtual(m->system, base, vaddr + amt - base, PROT_READ | PROT_WRITE, false) (ENOBUFS)
         PC 400078 segfault
         AX 0000000000000000  CX 0000000000000000  DX 0000000000000000  BX 0000000000000000
         SP 0000000000000000  BP 0000000000000000  SI 0000000000000000  DI 0000000000000000
         R8 0000000000000000  R9 0000000000000000 R10 0000000000000000 R11 0000000000000000
        R12 0000000000000000 R13 0000000000000000 R14 0000000000000000 R15 0000000000000000
         FS 0000000000000000  GS 0000000000000000 OPS 0                FLG ......
        third_party/cosmo/tinyhello.elf
        000000000000 000000400078 UNKNOWN
        <blink backtrace unavailable>
tlb_misses                       = 1
tlb_resets                       = 1
[1]   Abort trap (core dumped) o/blink/blink third_party/cosmo/tinyhello.elf
```

jart:
Try writing a C program locally on the system to confirm. Something like:

```c
TEST(mprotect, loosenRestriction) {
  char *p;
  ASSERT_NE(MAP_FAILED, (p = mmap(0, 1, PROT_READ | PROT_WRITE,
                                  MAP_ANONYMOUS | MAP_PRIVATE, -1, 0)));
  ASSERT_SYS(0, 0, mprotect(p, 1, PROT_READ | PROT_EXEC));
  ASSERT_SYS(0, 0, munmap(p, 1));
}
```

bjia56:
This is probably relevant:
```
vmactions# sysctl security.pax.mprotect.global
security.pax.mprotect.global = 1
```

bjia56:
Disabling PaX mprotect allows the original test to succeed

jart:
Is PaX enabled by default in the stock installation these days?

bjia56:
According to NetBSD 8.0's release notes, PaX mprotect is enabled by default on x86_64 and arm64: https://www.netbsd.org/releases/formal-8/NetBSD-8.0.html
I don't see any notes about PaX in subsequent release notes other than bugfixes.

bjia56:
It appears that PaX mprotect can be disabled on binaries with paxctl: https://man.netbsd.org/paxctl.8

bjia56:
For my use case, running `paxctl +m` on the compiled blink binary before adding it with apelink is sufficient. I'll leave it up to you if you want this PR in or not.

jart:
Thanks for the update. I'd rather not make it default to RWX memory on NetBSD because it has a WIN32-like memory restriction feature. Since there'll probably be another feature lurking in NetBSD somewhere that demands a W^X invariant. Appreciate you taking the time to send the change though! Glad you're unblocked.

jart:
Actually on second thought, having just noticed the comment that PAX is enabled by default in the stock install, what would you recommend? Would it be possible to configure the Blink build system to set the flag on the binary automatically for our users?

bjia56:
I think a short-term fix would be to add `paxctl +m` to the build system so binaries work, however it's a bit sketchy to disable a security feature on Blink entirely (since it could, as a VM, be used to run untrusted x86_64 binaries). A more permanent fix could be to identify the problematic parts and use [mremap](https://man.netbsd.org/mremap.2) with MAP_REMAPDUP to duplicate mappings when more permissions are required. Unfortunately, I am not as familiar with Blink internals at this point to make the latter change.

jart (approved):
You're right that we should respect the security blankets each platform
puts in place. This change does a better job at that, since it disables
PAX surgically, whereas `paxctl -m` would disable PAX systemically.

> since it could, as a VM, be used to run untrusted x86_64 binaries

Stock open source Blink shouldn't be thought of as a security layer for
untrusted binaries. Blink aims to ensure programs work above all, so
that people have a good experience using the software. That means Blink
lets your programs do what they want to do. *However* Blink is designed
with the security use case in mind. It should take minimal eng effort to
modify blink/syscall.c to enforce whatever security policy you want. If
you do that, then Blink can potentially offer you a much stronger layer
of hardware/system isolation than alternatives like KVM, gVisor, etc.
Especially if you disable JIT and have a SECCOMP policy too. We also
really should add soft floating point to make it even more extreme.

